PR #2, opened 2023-03-14
Severity rank is computed

lawrencejones:
As if ommitted, it is calculated for us. It may also change underneath is severities are reordered.


lawrencejones:
Tests are against the same account and collided on names, but thankfully we're good here (ran them locally).

Going to merge this and cut v1.0.1.

PR #17, opened 2023-03-29
Catalog entry attribute values are sets

lawrencejones:
As the order with which you specify an attribute is not relevant when we perform an update.


samstarling (approved):
Wonderful!

PR #19, opened 2023-04-11
incident_catalog_entries

lawrencejones:
Introduce a resource that can manage entries in bulk, which:

1. Solves terraform performance issues when dealing with a lot of entries
2. Reduces pressure on our API when working with entries

To achieve this we've forked the terraform sdk framework to improve performance when logging, which is necessary to avoid massive plan times on large nested objects.

Relevant links are:
- PR for logging change (https://github.com/hashicorp/terraform-plugin-framework/pull/722)
- Issue in upstream (https://github.com/hashicorp/terraform-plugin-framework/issues/721)


samstarling (approved):
This all looks good to me, a relative Terraform newbie. The README was very nice to read. Do you think there's value in having both "entry" and "entries"? To rephrase: it wasn't immediately clear to me the advantage of always just using this new "plural" provider/resource.

lawrencejones:
On the singular resource, I think there's little point removing it because:
1. Most people will come to this via official guidance/tutorial we'll write, which will highlight when you should use one over the over
2. It offers the ability to add a single entry or entries without external IDs without removing all entries that are not managed by terraform, so there is a comparative functionality difference

I'll add a similar doc comment to the singular resource as we have for the plural that gives advice on when/when not to use it.

PR #35, opened 2023-09-19
Don't ever put an empty array into ArrayValue

paprikati:


waltfy:
@paprikati I'm going to close this as it has been superseded by #36

PR #36, opened 2023-09-20
Catalog handles omitted empty lists

lawrencejones:
If we try creating a catalog entry with an array attribute value giving:
```json
{ "array_value": null }
```

The response from the API will omit the attribute. Similarly, this payload:
```json
{ "array_value": [] }
```

Will also result in an empty attribute response. The problem is that if your terraform code had the array_value attribute set to either null or empty list, terraform will remember the distinction, and when the API returns a list of attribute values where there is no value, terraform will perceive this as a state difference and die with the following error:

```

When applying changes to incident_catalog_entries.Service, provider
"provider[\"registry.terraform.io/incident-io/incident\"]" produced an
unexpected new value:
.entries["attr"].attribute_values["01H7Y6V5P9DP8HMCF2FN48SM6M"].array_value:
was null, but now cty.ListValEmpty(cty.String).
```

To handle this, we have changed how we build the terraform state from the API response so that when we think it's possible that the API omitted the attribute, we check our state and use the value in the state to ensure terraform is happy.

### DIFF
--- a/internal/provider/incident_catalog_entries_resource_test.go
+++ b/internal/provider/incident_catalog_entries_resource_test.go
@@ -22,11 +22,13 @@ func TestAccIncidentCatalogEntriesResource(t *testing.T) {
 						Name:        "One",
 						ExternalID:  "one",
 						Description: "This is the first entry",
+						ArrayValue:  "null",
 					},
 					{
 						Name:        "Two",
 						ExternalID:  "two",
 						Description: "This is the second entry",
+						ArrayValue:  "[]",
 					},
 				}),
 				Check: resource.ComposeAggregateTestCheckFunc(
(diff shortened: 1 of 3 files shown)


internal/provider/incident_catalog_entries_resource_test.go
```diff
@@ -22,45 +22,51 @@ func TestAccIncidentCatalogEntriesResource(t *testing.T) {
...
-						Description: "This is the third entry",
-					},
-				}),
-				Check: resource.ComposeAggregateTestCheckFunc(
-					resource.TestCheckResourceAttr(
-						"incident_catalog_entries.example", "entries.two.name", "Three"),
+				// resource.TestCheckResourceAttr(
+				// 	"incident_catalog_entries.example", "entries.one.name", "One"),
+				// resource.TestCheckResourceAttr(
+				// 	"incident_catalog_entries.example", "entries.two.name", "Two"),
```
waltfy, line 38: Bring this back?

internal/provider/incident_catalog_entries_resource_test.go
```diff
@@ -22,45 +22,51 @@ func TestAccIncidentCatalogEntriesResource(t *testing.T) {
...
+							ArrayValue:  "null",
+						},
+						{
+							Name:        "Three",
+							ExternalID:  "two",
+							Description: "This is the third entry",
+							ArrayValue:  "[]",
+						},
+					}),
+					Check: resource.ComposeAggregateTestCheckFunc(
```
waltfy, line 64: Bring this back?

waltfy (approved):
Looks good, let's cut a release and make sure we broadcast this bug so people can upgrade their providers?

PR #37, opened 2023-10-09
Move terraform to v2 custom field + incident role endpoints

paprikati:
We are deprecating a number of fields from the CustomFieldsV1 API (required, show_before_closure, show_before_creation, show_before_update, show_in_announcement_post). and IncidentRolesV1 API (required). These will now be controlled via 'Incident Forms' which (for now) will only be available via the web dashboard.

There's also a few other bugfixes, described in each commit:

1008d5c **Regenerate client from latest public API schema** 
 
0682230 **Point code at renamed types from the new openapi schema** 
 We renamed a number of catalog public API types to 'Engine' public API types, as part of making our internal APIs more consistent. This updates all the callsites to use the new equivalent types

fc6c042 **Remove SemanticType from catalog types** 
 This has never been used by our application, so we've decided to remove it from the provider as we have no plans to use it.

fbd9e2b **Move to CustomFieldsV2 API** 
 We are deprecating a number of fields from the CustomFieldsV1 API (required, show_before_closure, show_before_creation, show_before_update, show_in_announcement_post). These will now be controlled via 'Incident Forms' which (for now) will only be available via the web dashboard.
This will enable users to have much more control over the way they configure their incident forms.

c12a7c0 **Move to IncidentRolesV2 API** 
 We are deprecating the `required` field from the IncidentRolesV1 API.
This will now be controlled via 'Incident Forms' which (for now) will only be available via the web dashboard.
This will enable users to have much more control over the way they configure their incident forms.

28caaa1 **Update changelog** 
 
185b15e **Fix test: look for correct schema tag**


lawrencejones (commented):
Approach looks good, few questions though:
- What is the upgrade path for people using these fields? We've got a few people who'll be relying on these so how do we advise them to handle the change?
- Also just simply what happens if you have a terraform project using the old provider and then upgrade to this one, how does it handle the state/attributes changing.

paprikati:
> What is the upgrade path for people using these fields? We've got a few people who'll be relying on these so how do we advise them to handle the change?

Added to the changelog, you can simply delete the deprecated fields

> Also just simply what happens if you have a terraform project using the old provider and then upgrade to this one, how does it handle the state/attributes changing.

Tested locally, verified that it handles the removed fields gracefully

lawrencejones (approved):

PR #40, opened 2023-12-14
docs fixes

ksatirli:
Hey folks 👋🏼,

While testing the provider, I came across a few things that I think are worth fixing.

Other than the (job-induced) nits, there are two main points:

1.) `required` is not part of `incident_incident_role`, so the example code doesn't work

2.) `incident_severity` references `Major`, which is (as far as I can see) a pre-existing severity level.


lawrencejones (approved):
Thanks for this!

PR #41, opened 2024-01-10
PR-254 Support setting the catalog type_name explicitly

isaacseymour:
This means that the catalog type can be referred to as `Custom["Team"]` rather than `Custom["012346325AGDFADY612946AN"]`, which is nice!


samstarling (approved):
Looks very sensible to me, a nice addition: thank you!

PR #46, opened 2024-03-26
Support backlink attributes

paprikati:
We'll now enable you to create backlink attributes via terraform, by specifying the `backlink_attribute` in the resource definition.


internal/provider/incident_catalog_entries_resource.go
```diff
@@ -399,7 +399,7 @@ func (r *IncidentCatalogEntriesResource) getEntries(ctx context.Context, catalog
 	for {
 		result, err := r.client.CatalogV2ListEntriesWithResponse(ctx, &client.CatalogV2ListEntriesParams{
 			CatalogTypeId: catalogTypeID,
-			PageSize:      lo.ToPtr(int(250)),
+			PageSize:      lo.ToPtr(int64(250)),
```
benjiAtIncident, line 402: Just for my own curiosity, what prompted this?

benjiAtIncident (approved):
The logic here all looks legit to me

internal/provider/incident_catalog_entries_resource.go
paprikati: It's because we've updated our openapi schema. My guess is that we've changed the type in our own API definition, which is why the openapi generator has reflected it here

PR #54, opened 2024-04-19
Add initial workflow resource

louisheath:
To come later; steps, expressions, condition groups, polish


louisheath:
For client gen I ran
```
rm -rf internal/client/client.gen.go; oapi-codegen --generate types,client --package client --o internal/client/client.gen.go internal/apischema/openapi3.json
```

internal/provider/incident_workflow_resource.go
```diff
@@ -0,0 +1,196 @@
...
+	return &IncidentWorkflowResource{}
+}
+
+type IncidentWorkflowResourceModel struct {
+	ID               types.String `tfsdk:"id"`
+	Name             types.String `tfsdk:"name"`
+	Folder           types.String `tfsdk:"folder"`
+	Trigger          types.String `tfsdk:"trigger"`
+	TerraformRepoURL types.String `tfsdk:"terraform_repo_url"`
+}
```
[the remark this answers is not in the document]
louisheath: (It's lightweight to begin, I had some [more coded up which I deleted](https://github.com/incident-io/terraform-provider-incident/pull/54/commits/3deacb6fe638b14b9c06b8b6f67c554771771e32#diff-71176964b924e4fba8771ef63364bcf8f397f57ab4ac7bf46e054719b6ac0cc1), and which might be useful later)

internal/provider/incident_workflow_resource.go
```diff
@@ -0,0 +1,210 @@
...
+	resp.Diagnostics.Append(req.State.Get(ctx, &state)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	var data *IncidentWorkflowResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
```
louisheath, line 113: read from state to find the `ID`, but from the _plan_ to find the payload. we could also consider merging non-empty fields from the plan onto the state and using the merged payload in the JSON request body.. this isn't set up for now

internal/provider/incident_workflow_resource.go
```diff
@@ -0,0 +1,210 @@
...
+				Required: true,
+			},
+			"folder": schema.StringAttribute{
+				Optional: true,
+			},
+			"trigger": schema.StringAttribute{
+				Required: true,
+			},
+			"terraform_repo_url": schema.StringAttribute{
+				Optional: true,
```
paprikati, line 58: I wonder if we should make this required? This is what's going to control the stuff which blocks you from editing it on our side, so I think we'll always want you to have it

internal/provider/incident_workflow_resource.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	var data *IncidentWorkflowResourceModel
+	resp.Diagnostics.Append(req.Plan.Get(ctx, &data)...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	payload := client.WorkflowsV2UpdateWorkflowJSONRequestBody{
+		Workflow: client.WorkflowPayload{
```
paprikati, line 116: I think you're missing Folder here

internal/provider/incident_workflow_resource.go
```diff
@@ -0,0 +1,210 @@
...
+			"id": schema.StringAttribute{
+				Computed: true,
+			},
+			"name": schema.StringAttribute{
+				Required: true,
+			},
+			"folder": schema.StringAttribute{
+				Optional: true,
+			},
+			"trigger": schema.StringAttribute{
```
paprikati, line 54: This can't be changed once set (well, it requires a delete), we can encode that using a plan modifier:
```
PlanModifiers: []planmodifier.String{
				stringplanmodifier.RequiresReplace(),
			},
```

paprikati (approved):
Couple of comments. Your tests won't pass until the public API is on staging (as the acceptance tests run against staging), I think it's fine to merge this into the feature branch as-is so we can keep incrementally shipping

internal/provider/incident_workflow_resource.go
louisheath: There's no folder on the payload ! I think it's a different endpoint .. which may make things non-atomic if we allow folder updates

paprikati: I think we should just fix that, and have the folder also in the main update payload, and the frontend can just use the utility endpoint for its own stuff. CC @benjiAtIncident

louisheath: https://linear.app/incident-io/issue/CAT-272

PR #56, opened 2024-04-23
Add steps to workflow resource

louisheath:


internal/provider/incident_workflow_resource.go
```diff
@@ -83,23 +109,7 @@ func (r *IncidentWorkflowResource) Schema(ctx context.Context, req resource.Sche
 									"operation": schema.StringAttribute{
 										Required: true,
 									},
-									"param_bindings": schema.SetNestedAttribute{
```
louisheath, line 86: `SetNestedAttribute` is "a Set of Nested Attibutes", not "Set these Nested Attributes"

which completely explains the lack of ordering

internal/provider/incident_workflow_resource.go
```diff
@@ -51,6 +59,24 @@ func (r *IncidentWorkflowResource) Schema(ctx context.Context, req resource.Sche
...
+						Attributes: paramBindingValueAttributes,
+					},
+				},
+				"value": schema.SingleNestedAttribute{
+					Optional:   true,
+					Attributes: paramBindingValueAttributes,
+				},
+			},
+		},
+	}
```
louisheath, line 78: feels good to flatten this nest

internal/provider/incident_workflow_resource.go
```diff
@@ -283,6 +311,21 @@ func (r *IncidentWorkflowResource) buildConditionGroups(groups []client.Expressi
...
 
+func (r *IncidentWorkflowResource) buildSteps(steps []client.StepConfig) []IncidentWorkflowStep {
+	var out []IncidentWorkflowStep
+
+	for _, s := range steps {
+		out = append(out, IncidentWorkflowStep{
+			ForEach:       types.StringPointerValue(s.ForEach),
+			Id:            types.StringValue(s.Id),
+			Name:          types.StringValue(s.Name),
+			ParamBindings: r.buildParamBindings(s.ParamBindings),
```
louisheath, line 322: and feels good to re-use these straight away

internal/provider/incident_workflow_resource_test.go
```diff
@@ -51,47 +55,78 @@ func TestAccIncidentWorkflowResource(t *testing.T) {
...
 resource "incident_workflow" "example" {
-  name               = {{ quote .Name }}
-  trigger            = "incident.updated"
-  terraform_repo_url = "https://github.com/incident-io/test"
-  condition_groups = [
-	{
-	  conditions = [
-		 {
-			operation = "one_of"
+	name               = {{ quote .Name }}
```
louisheath, line 82: fix indentation and add steps

docs/resources/workflow.md
```diff
@@ -72,3 +73,48 @@ Optional:
...
+### Nested Schema for `steps`
+
+Required:
+
+- `name` (String)
+- `param_bindings` (Attributes List) (see [below for nested schema](#nestedatt--steps--param_bindings))
+
+Optional:
+
+- `for_each` (String)
```
paprikati, line 89: We should definitely document these better, can we make sure we have a card to make these really nice?

internal/provider/incident_workflow_resource.go
paprikati: Can you explain what you mean here? i'm not quite sure!

internal/provider/incident_workflow_resource.go
paprikati: OH right this makes sense

internal/provider/incident_workflow_resource_test.go
```diff
@@ -51,47 +55,78 @@ func TestAccIncidentWorkflowResource(t *testing.T) {
...
 			},
+			// Update step and check new state
+			{
+				Config: testAccIncidentWorkflowResourceConfig(&workflowTemplateOverrides{
+					StepSlackMessage: "Don't forget to write a postmortem!",
+				}),
+				Check: resource.ComposeAggregateTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"incident_workflow.example", "steps.0.param_bindings.1.value.literal",
+						"{\"content\":[{\"content\":[{\"text\":\"Don't forget to write a postmortem!\",\"type\":\"text\"}],\"type\":\"paragraph\"}],\"type\":\"doc\"}"),
```
paprikati, line 66: I think we probably want our example to be something that doesn't include templated text? Just cause it's ugly and kinda distracting in our docs, I think

paprikati (approved):
I think lets keep going to our MVP, but might come back and change this structure as per the slack convo!

docs/resources/workflow.md
louisheath: we haven't linked to the API docs just yet. I found it very fiddly due to the number of subtly differently named payload types

I've included it in a polish ticket which also covers ironing out the other fields (bool flags etc) https://linear.app/incident-io/issue/CAT-275

internal/provider/incident_workflow_resource.go
louisheath: it was just very deeply nested >>>>>

internal/provider/incident_workflow_resource.go
louisheath: yea I was getting workflow validation errors because my params were in the wrong order

docs/resources/workflow.md
paprikati: tbh I would just write it out, I think using the docs is more effort than it's worth here

PR #57, opened 2024-04-23
Louis/cat 250 exp

louisheath:
There's a lot more payload to this one ..


internal/provider/incident_workflow_resource.go
```diff
@@ -400,27 +494,108 @@ func toPayloadParamBindings(pbs []IncidentEngineParamBinding) []client.EnginePar
...
 
-	return paramBindings
+	return out
+}
+
+func toPayloadReturns(returns IncidentEngineReturnsMeta) client.ReturnsMetaV2 {
+	return client.ReturnsMetaV2{
+		Array: returns.Array.ValueBool(),
+		Type:  returns.Type.ValueString(),
+	}
```
louisheath, line 600: converters taking types `IncidentEngine*` could also live in engine.go

PR #58, opened 2024-04-30
Add incident_io_user Data Source

rorymalcolm:
- Adds a `data` source, `incident_io_user`

- This lets users fetch information about incident.io users in the Terraform provider

- Have explicitly decided _not_ to add acceptance tests here - data sources are obviously materially different from our resource tests!


internal/apischema/openapi3.json
```diff
@@ -783,6 +783,7 @@
                 "github_pull_request",
                 "gitlab_merge_request",
                 "sentry_issue",
+                "jira_issue",
```
henryjcee, line 786: is this intentional

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+			return
+		}
+		user = &result.JSON200.User
+	} else if !data.Email.IsNull() {
+		result, err := i.client.UsersV2ListWithResponse(ctx, &client.UsersV2ListParams{
+			Email: data.Email.ValueStringPointer(),
+		})
+		if err == nil && result.StatusCode() >= 400 {
+			err = fmt.Errorf(string(result.Body))
+		}
```
henryjcee, line 96: In what circumstances do we get a nil error and 4++ response?

henryjcee (approved):
One question, otherwise this looks sensible.

internal/provider/incident_user_data_source.go
rorymalcolm: Same as what we do elsewhere in the provider!

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+func NewIncidentUserDataSource() datasource.DataSource {
+	return &IncidentUserDataSource{}
+}
+
+type IncidentUserDataSource struct {
+	client *client.ClientWithResponses
+}
+
+type IncidentUserDataSourceModel struct {
+	BaseRole    *RBACRole    `tfsdk:"base_role" json:"base_role"`
```
paprikati, line 29: Do we care about roles here? If not, should we just slim this down to what we actually need to make our life simpler?

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+	CustomRoles []*RBACRole  `tfsdk:"custom_roles" json:"custom_roles"`
+	Email       types.String `tfsdk:"email" json:"email"`
+	ID          types.String `tfsdk:"id" json:"id"`
+	Name        types.String `tfsdk:"name" json:"name"`
+	Role        types.String `tfsdk:"role" json:"role"`
+	SlackUserId types.String `tfsdk:"slack_user_id" json:"slack_user_id"`
```
paprikati, line 35:
```suggestion
	SlackUserID types.String `tfsdk:"slack_user_id" json:"slack_user_id"`
```
would be more go-y

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+}
+
+type IncidentUserRequest struct {
+	ID types.String `tfsdk:"id"`
+}
+
+type RBACRole struct {
+	Description types.String `tfsdk:"description"`
+	Id          types.String `tfsdk:"id"`
```
paprikati, line 44:
```suggestion
	ID          types.String `tfsdk:"id"`
```
the inconsistency is killing me

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+	i.client = client
+}
+
+func (i *IncidentUserDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_user"
+}
+
+func (i *IncidentUserDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var data IncidentUserDataSourceModel
+	req.Config.Get(ctx, &data)
```
paprikati, line 73: Do we want to do the `resp.Diagnostics.Append` dance here?

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+		if err == nil && result.StatusCode() >= 400 {
+			err = fmt.Errorf(string(result.Body))
+		}
+		if err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read user, got error: %s", err))
+			return
+		}
+		user = &result.JSON200.User
+	} else if !data.Email.IsNull() {
+		result, err := i.client.UsersV2ListWithResponse(ctx, &client.UsersV2ListParams{
```
paprikati, line 91: I quite dislike this vs. having a lookup endpoint where you can provide an email, but I think we've had this for ages so I'll code

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+			Email: data.Email.ValueStringPointer(),
+		})
+		if err == nil && result.StatusCode() >= 400 {
+			err = fmt.Errorf(string(result.Body))
+		}
+		if err != nil {
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read user, got error: %s", err))
+			return
+		}
+		if len(result.JSON200.Users) == 0 {
```
paprikati, line 101: I think we might want to log.Debug if we get more than one user?

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,207 @@
...
+			resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read user, got error: %s", "User not found"))
+			return
+		}
+		user = &result.JSON200.Users[0]
```
paprikati, line 121: same here, should never happen, but lets scream if we have >1

internal/provider/incident_user_data_source.go
rorymalcolm: I just added it today lol

paprikati: hmm. I don't love this in our public API vs. an explicit lookup endpoint. One for #api-design? Not a hill I'll die on

internal/provider/incident_user_data_source.go
rorymalcolm: I think it's much easier if this is a faithful representation of what the API has in it

paprikati: But as soon as we add things to the API, it won't be, right?

paprikati: Like, we can always add these roles later if someone needs them, but all we're doing by including them is making it harder for us to change the representation of the roles later, I think (e.g. changing role slugs)

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func (i *IncidentUserDataSource) Metadata(ctx context.Context, req datasource.MetadataRequest, resp *datasource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_user"
+}
+
+func (i *IncidentUserDataSource) Read(ctx context.Context, req datasource.ReadRequest, resp *datasource.ReadResponse) {
+	var data IncidentUserDataSourceModel
+	req.Config.Get(ctx, &data)
+	resp.Diagnostics.Append(req.Config.Get(ctx, &data)...)
```
paprikati, line 70: you've got this twice now!

internal/provider/incident_user_data_source.go
```diff
@@ -0,0 +1,162 @@
...
+	ID          types.String `tfsdk:"id" json:"id"`
+	Name        types.String `tfsdk:"name" json:"name"`
+	SlackUserID types.String `tfsdk:"slack_user_id" json:"slack_user_id"`
+}
+
+type IncidentUserRequest struct {
+	ID types.String `tfsdk:"id"`
+}
+
+type RBACRole struct {
```
paprikati, line 38: ✂️

docs/data-sources/user.md
```diff
@@ -0,0 +1,55 @@
...
+## Schema
+
+### Optional
+
+- `email` (String)
+- `slack_user_id` (String)
+
+### Read-Only
+
+- `base_role` (Attributes) (see [below for nested schema](#nestedatt--base_role))
```
paprikati, line 28: need to regen

paprikati (approved):
once you've removed those couple of bits!

PR #62, opened 2024-05-08
`incident_schedule` example Terraform

rorymalcolm:
- Provide an example of how best to use the incident.io schedule resource


examples/resources/incident_schedule/resource.tf
```diff
@@ -0,0 +1,76 @@
...
+}
+
+data "incident_user" "roryb" {
+  email = "rory@incident.io"
+}
+
+resource "incident_schedule" "testing_my_provider" {
+  name = "Testing Terraform"
+
+  // from tz database
```
lawrencejones, line 16: Shall we expand this? Presumably this is an ISO format we can reference?

examples/resources/incident_schedule/resource.tf
```diff
@@ -0,0 +1,76 @@
...
+
+data "incident_user" "martha" {
+  email = "martha@incident.io"
+}
+
+data "incident_user" "roryb" {
+  email = "rory@incident.io"
+}
+
+resource "incident_schedule" "testing_my_provider" {
```
lawrencejones, line 13:
```suggestion
resource "incident_schedule" "primary_on_call" {
```

examples/resources/incident_schedule/resource.tf
```diff
@@ -0,0 +1,76 @@
...
+  name = "Testing Terraform"
```
lawrencejones, line 14:
```suggestion
  name = "Primary On-call"
```

examples/resources/incident_schedule/resource.tf
```diff
@@ -0,0 +1,76 @@
...
+
+  // from tz database
+  timezone = "Europe/London"
+
+  rotations = [{
+    // A string ID for a schedule rotation
+    id                = "01HPFH8T92MPGSQS5C1SPAF4V0"
```
lawrencejones, line 21: How do I pick this value? Presumably we're not asking people to ULID generate it, right?

Is it not the case that most people will end up calling this something they create by hand like:
```
id = "primary"
name = "Primary"
```

examples/resources/incident_schedule/resource.tf
rorymalcolm: There's no ISO format afaik - `tz` is the scientific term! https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

lawrencejones (approved):

PR #64, opened 2024-05-13
Add remaining fields to workflows model

paprikati:
This updates to our latest openapi schema, and also implements all the 'other' fields that we hadn't touched yet.

It also changes expressions to be a map of `reference: expression` so that we don't get in trouble when the expressions don't come back in a consistent order.

### DIFF
--- a/internal/provider/engine.go
+++ b/internal/provider/engine.go
@@ -29,13 +29,14 @@ type IncidentEngineParamBindingValue struct {
 	Reference types.String `tfsdk:"reference"`
 }
 
-type IncidentEngineExpressions map[string]IncidentEngineExpression
+type IncidentEngineExpressions []IncidentEngineExpression
 
 type IncidentEngineExpression struct {
 	ElseBranch    *IncidentEngineElseBranch           `tfsdk:"else_branch"`
 	ID            types.String                        `tfsdk:"id"`
 	Label         types.String                        `tfsdk:"label"`
 	Operations    []IncidentEngineExpressionOperation `tfsdk:"operations"`
+	Reference     types.String                        `tfsdk:"reference"`
 	RootReference types.String                        `tfsdk:"root_reference"`
 }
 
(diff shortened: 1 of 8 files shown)


internal/provider/engine.go
```diff
@@ -73,3 +79,147 @@ type IncidentEngineExpressionParseOpts struct {
 	Returns IncidentEngineReturnsMeta `tfsdk:"returns"`
 	Source  types.String              `tfsdk:"source"`
 }
+
+// Attributes
```
paprikati, line 83: These were defined inline, I've pulled them out so it's easier to read and we can re-use in other resources going forwards

internal/provider/incident_workflow_resource.go
```diff
@@ -21,23 +23,34 @@ var (
...
+	IncludePrivateIncidents types.Bool                    `tfsdk:"include_private_incidents"`
+	ContinueOnStepError     types.Bool                    `tfsdk:"continue_on_step_error"`
+	Delay                   *IncidentWorkflowDelay        `tfsdk:"delay"`
+	RunsOnIncidents         types.String                  `tfsdk:"runs_on_incidents"`
+	RunsOnIncidentModes     []types.String                `tfsdk:"runs_on_incident_modes"`
+	State                   types.String                  `tfsdk:"state"`
+}
+
+// 	IncludePrivateIncidents bool `json:"include_private_incidents"`
+// 	IncludeRetrospectiveIncidents bool `json:"include_retrospective_incidents"`
```
louisheath, line 53: 🔪

internal/provider/incident_workflow_resource.go
```diff
@@ -401,7 +384,7 @@ func (r *IncidentWorkflowResource) buildConditionGroups(groups []client.Expressi
 }
 
 func (r *IncidentWorkflowResource) buildConditions(conditions []client.ConditionV2) []IncidentEngineCondition {
-	var out []IncidentEngineCondition
+	out := []IncidentEngineCondition{}
```
louisheath, line 387: so that we always end up with an empty json array , to not hit `required` validation errs ?

louisheath (approved):

internal/provider/incident_workflow_resource.go
paprikati: Yep exactly that.

paprikati:
@louisheath this is ready for re-review!

internal/provider/incident_workflow_resource.go
```diff
@@ -287,19 +212,41 @@ func (r *IncidentWorkflowResource) Update(ctx context.Context, req resource.Upda
...
 
+	onceFor := []string{}
+	for _, v := range data.OnceFor {
+		onceFor = append(onceFor, v.ValueString())
+	}
+
+	runsOnIncidentModes := []client.UpdateWorkflowRequestBodyRunsOnIncidentModes{}
+	for _, v := range data.RunsOnIncidentModes {
+		runsOnIncidentModes = append(runsOnIncidentModes, client.UpdateWorkflowRequestBodyRunsOnIncidentModes(v.ValueString()))
+	}
```
louisheath, line 223: nit nit we _could_ extract these two conversions into funcs and re-use them across update and create.

louisheath (approved):

internal/provider/incident_workflow_resource.go
paprikati: You can't, because the create and update types are different sadly

PR #65, opened 2024-05-13
Add missing fields to workflow tests

benjiAtIncident:
This PR adds a number of missing fields to the struct, and adds the `Id` field to expressions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.dll
 *.exe
+.env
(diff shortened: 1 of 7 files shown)


benjiAtIncident:
The failing tests are passing locally - I'm assuming they're failing here because they're 404-ing on the api since the FF isn't on?

paprikati (approved):

PR #68, opened 2024-05-15
Claim resources when you import them

paprikati:
This means that we set managed_by=terraform against these resources when you import them, so the nice purple callout appears and we stop you editing it from the UI

### DIFF
--- a/internal/provider/incident_schedule_resource.go
+++ b/internal/provider/incident_schedule_resource.go
@@ -308,6 +308,7 @@ func (r *IncidentScheduleResource) Delete(ctx context.Context, req resource.Dele
 }
 
 func (r *IncidentScheduleResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	claimResource(ctx, r.client, req, resp, client.ManagedResourceV2ResourceTypeSchedule, r.terraformVersion)
 	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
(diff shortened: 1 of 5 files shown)


internal/provider/incident_schedule_resource.go
```diff
@@ -308,6 +308,7 @@ func (r *IncidentScheduleResource) Delete(ctx context.Context, req resource.Dele
 }
 
 func (r *IncidentScheduleResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	claimResource(ctx, r.client, req, resp, client.ManagedResourceV2ResourceTypeWorkflow, r.terraformVersion)
```
rorymalcolm, line 311: This should be schedule!

internal/provider/incident_workflow_resource.go
```diff
@@ -562,7 +563,7 @@ func toPayloadSteps(steps []IncidentWorkflowStep) []client.StepConfigPayload {
 	for _, step := range steps {
 		out = append(out, client.StepConfigPayload{
 			ForEach:       step.ForEach.ValueStringPointer(),
-			Id:            step.ID.ValueStringPointer(),
```
rorymalcolm, line 565: I take it we made these non-computed or equivalent?

paprikati: Yeah exactly - this change is because I re-genned the client which now has them as required to match this

PR #75, opened 2024-05-24
data.incident_catalog_type

lawrencejones:
In case people want to reference catalog types by either their name or the type_name, this data resource allows them to do that.


docs/resources/schedule.md
pipt, line 1: Extraneous whitespace

docs/data-sources/catalog_type.md
```diff
@@ -0,0 +1,29 @@
...
+
+
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Optional
+
+- `name` (String) Name is the human readable name of this type
+- `type_name` (String) The type name of this catalog type, to be used when defining attributes. This is immutable once a CatalogType has been created. For non-externally sync types, it must follow the pattern Custom["SomeName "]
```
pipt, line 21: This should be `Custom["SomeName"]` without the trailing space

internal/provider/incident_catalog_type_data_source.go
```diff
@@ -0,0 +1,125 @@
...
+			"type_name": schema.StringAttribute{
+				MarkdownDescription: apischema.Docstring("CatalogV2CreateTypeRequestBody", "type_name"),
+				Optional:            true,
+				Computed:            true,
+			},
+			"description": schema.StringAttribute{
+				MarkdownDescription: apischema.Docstring("CatalogTypeV2ResponseBody", "description"),
+				Computed:            true,
+			},
+			"source_repo_url": schema.StringAttribute{
```
pipt, line 49: Is this needed? I know the API returns it at the moment, but we introduced managed resources in the backend a week or so ago, and if it's not needed by this change I'd avoid including it in case it causes us migration trouble in the future.

internal/provider/incident_catalog_type_data_source.go
```diff
@@ -0,0 +1,125 @@
...
+	if err == nil && result.StatusCode() >= 400 {
+		err = fmt.Errorf(string(result.Body))
+	}
+	if err != nil {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to read catalog types, got error: %s", err))
+		return
+	}
+
+	if data.Name.IsNull() && data.TypeName.IsNull() {
+		resp.Diagnostics.AddError("Client Error", fmt.Sprintf("Unable to find catalog type, got error: %s", "Either name or external_type must be set"))
```
pipt, line 96: Should `external_type` be `type_name`?

internal/provider/incident_catalog_type_data_source.go
```diff
@@ -0,0 +1,125 @@
...
+		return
+	}
+
+	catalogTypes := result.JSON200.CatalogTypes
+	if !data.Name.IsNull() {
+		catalogTypes = lo.Filter(catalogTypes, func(ct client.CatalogTypeV2, _ int) bool {
+			return ct.Name == data.Name.ValueString()
+		})
+	}
+	if !data.TypeName.IsNull() {
```
pipt, line 106: Should we document that we'll use `type_name` over `name` for the lookup if both are given?

pipt (approved):
A few comments, but looks good to me

internal/provider/incident_catalog_type_data_source.go
lawrencejones: it shoud

internal/provider/incident_catalog_type_data_source.go
lawrencejones: Do we not use both?

docs/data-sources/catalog_type.md
lawrencejones: Ohh, this is actually from our API docs 🤦 can't change here.

lawrencejones:
Gonna get this in and cut a new release.

internal/provider/incident_catalog_type_data_source.go
pipt: I meant that if they're both given, and they end up returning different catalog types, we'll return the type where the `type_name` matched rather than the `name`.

PR #85, opened 2024-06-17
incident_escalation_path

lawrencejones:
Support escalation paths. This includes a bump of our API schema which
requires us to change a number of resources, but should leave all other
resources functionality unchanged.


rorymalcolm (approved):
LGTM with one nit comment - well tested!

PR #86, opened 2024-06-27
Add support for categories

paprikati:
Introduced a new property `categories` on `resource_catalog_type` which you can use to make your catalog easier to navigate.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ build:
 # Clients
 ################################################################################
 
-.PHONY: internal/client
+.PHONY: internal/client/client.gen.go
 
 internal/client/client.gen.go:
 	rm -rf $@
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -26,7 +26,7 @@ build:
 # Clients
 ################################################################################
 
-.PHONY: internal/client
+.PHONY: internal/client/client.gen.go
```
paprikati, line 29: unrelated: without this, you have to manually delete the client before re-generating it

internal/apischema/openapi3.json
```diff
@@ -6049,56 +6079,6 @@
         }
       }
     },
-    "/v2/managed_resources": {
```
pipt, line 6052: Is this removal intentional?

pipt (approved):
One question about the managed resources API removal from the API json, but looks good :+1:

internal/apischema/openapi3.json
paprikati: Yes, I've copied over all the schemas which is what lojo was expecting when he made the change. We only read 'secret' but to make it harder to accidentally pull in the wrong schema we now have all of them in a single folder

PR #91, opened 2024-07-04
Add round robin and schedule types to escalation path resource

LucasBN:
Updates the escalation path terraform API to include the `schedule_mode` field and the `round_robin_config` field so that it has parity with the UI and the public API.


example/dev.tfrc
```diff
@@ -0,0 +1,7 @@
+provider_installation {
```
lawrencejones, line 1: @LucasBN ?

LucasBN: Sorry sir

rorymalcolm (approved):

PR #95, opened 2024-07-17
Retry 429 errors, respecting the Retry-After header

paprikati:
This means that if someone receives a 429, we'll retry after a sensible backoff, instead of the provider just stopping and failing.


internal/provider/provider.go
```diff
@@ -104,14 +106,30 @@ func (p *IncidentProvider) Configure(ctx context.Context, req provider.Configure
...
-	base := cleanhttp.DefaultClient()
-	base.Transport = &loghttp.Transport{
+	base := retryablehttp.NewClient()
+	base.RetryMax = 10
+	base.Backoff = func(min, max time.Duration, attemptNum int, httpResp *http.Response) time.Duration {
+		// Retry for rate limits and server errors.
+		if httpResp != nil && httpResp.StatusCode == http.StatusTooManyRequests {
+			retryAfter := httpResp.Header.Get("Retry-After")
+			if retryAfter != "" {
+				if retryAfterSeconds, err := time.ParseDuration(retryAfter + "s"); err == nil {
```
pipt, line 116: We return an RFC1123 timestamp rather than a duration, so this won't parse the `Retry-After` header correctly. I've got a WIP PR here that parses the header: https://github.com/incident-io/catalog-importer/pull/138

PR #98, opened 2024-07-18
Fix version number in changelog

pipt:


paprikati (approved):
Can we update to include the skipped versions?

PR #103, opened 2024-08-17
add incident role data source

tylerbrewer2:
Adds an `incident_role` data source


examples/data-sources/incident_catalog_type/data-source.tf
tylerbrewer2, line 1: Noticed the file was named incorrectly, so I fixed the file name and regenerated docs so the example is included in `docs/data-sources/catalog_type.md`

henryjcee:
Evening Tyler, thanks for the contribution. I've opened #105 that includes the changes in here, rebased on the major changes we've been making to our generated API that'll get merged as part of #104. I've made some changes to the acceptance tests so take a look and let me know what you think.

PR #104, opened 2024-08-22
Use our new openAPI specs

henryjcee:
Now using our lastest openAPI specs. This involves adding some more tedious type conversion between similar types (the spec hasn't really changed much, I think we just got lucky previously and only needed a handful of conversion functions) and  renaming some of the prop names in the acceptance tests.

### DIFF
--- a/internal/apischema/openapi.json
+++ b/internal/apischema/openapi.json
@@ -1233,10 +1233,7 @@
         "responses": {
           "200": {
             "description": "OK response.",
-            "schema": {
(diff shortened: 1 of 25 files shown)


henryjcee:
Bumped us to 1.22 and latest go linter in an effort to get the linter to stop complaining. Neither have worked and it's still complaining about non-const strings in the calls to the error package so I've disabled the linter for now. All tests are green.

paprikati (approved):

PR #105, opened 2024-08-22
Adds an incident role data source

henryjcee:
This builds on the downstream PR #103, rebases it on the new API changes in #104  and fixes it so that it works and acc tests are correct and pass.

#104 needs reviewing and merging before this goes in but I'm reasonably sure that this is now correct and has relevant tests.


internal/provider/incident_role_data_source.go
```diff
@@ -0,0 +1,118 @@
...
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				MarkdownDescription: apischema.Docstring("IncidentRoleV2", "id"),
+				Required:            true,
+			},
+			"name": schema.StringAttribute{
+				MarkdownDescription: apischema.Docstring("IncidentRoleV2", "name"),
+				Computed:            true,
+			},
+			"description": schema.StringAttribute{
```
paprikati, line 70: I'm gonna slim this down, I don't think this other information is actually useful to people and it just makes any migrations we might want to do a little harder (and makes the docs less clear imo)

paprikati (approved):

PR #114, opened 2024-10-23
Add `external_id` to `resource_catalog_entry`

rorymalcolm:
- This will allow users to set an external ID using this resource

### DIFF
--- a/examples/resources/incident_catalog_entry/resource.tf
+++ b/examples/resources/incident_catalog_entry/resource.tf
@@ -1,3 +1,23 @@
+locals {
+  service_tiers = [
+    {
+      name        = "tier_1"
+      description = "Critical customer-facing services"
+      external_id = "service-tier-1"
+    },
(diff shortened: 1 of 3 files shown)


examples/resources/incident_catalog_entry/resource.tf
```diff
@@ -32,6 +32,8 @@ resource "incident_catalog_entry" "service_tier" {
 
   name = each.value.name
 
+  external_id = each.value.name
```
paprikati, line 35: I think this is a little confusing and we should make the example better with a distinct name and id

paprikati (approved):
assuming you've tested it, then yes. @pipt do you think this should be optional or mandatory?

rorymalcolm:
> assuming you've tested it, then yes. @pipt do you think this should be optional or mandatory?

Tested locally and w/ acceptance tests!

PR #117, opened 2024-10-28
Handle validation errors for 204 endpoints

lawrencejones:
When the successful response to a client request is 204, it is possible that we'll parse an error response into successful (empty) result and not detect that it is an error.

This change adds the validation error handler we've used in other tools such as the catalog-importer so that we detect errors properly. This was a report from a customer who saw that Terraform trying to delete a catalog type resulted in a validation error from our servers but the catalog type was never deleted (because the endpoint had 422'd).

Confirmed this behaves as expected locally:
Screenshot 2024-10-28 at 16 44 43

### DIFF
--- a/internal/client/client.go
+++ b/internal/client/client.go
@@ -1,0 +1,121 @@
+package client
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/deepmap/oapi-codegen/pkg/securityprovider"
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/pkg/errors"
+)
+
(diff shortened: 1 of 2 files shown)


internal/client/client.go
```diff
@@ -0,0 +1,121 @@
...
+	"io"
+	"net/http"
+	"time"
+
+	"github.com/deepmap/oapi-codegen/pkg/securityprovider"
+	"github.com/hashicorp/go-retryablehttp"
+	"github.com/pkg/errors"
+)
+
+func New(ctx context.Context, apiKey, apiEndpoint, version string, opts ...ClientOption) (*ClientWithResponses, error) {
```
lawrencejones, line 15: This looks much more like our catalog-importer client now: https://github.com/incident-io/catalog-importer/blob/master/client/client.go

henryjcee (approved):
lgtm, changes look very sensible.

PR #121, opened 2025-01-22
Update API schema and docs

samstarling:
Moves us to the latest version of our API schema and docs, including tweaks to documentation about support for catalog-backed custom fields.


docs/resources/custom_field.md
```diff
@@ -9,8 +9,6 @@ description: |-
   your own data needs.
   Each field has a type:
   Single-select, single value selected from a predefined list of options (e.g. Detection Method)Multi-select, as above but you can pick more than one option (e.g. Affected Teams)Text, freeform text field (e.g. Customer ID)Link, link URL that is synced to Slack bookmarks on the incident channel (e.g. External Status Page)Number, integer or fractional numbers (e.g. # Customers Affected)
```
isaacseymour, line 11: this looks mangled (but historic)

isaacseymour (approved):

docs/resources/custom_field.md
samstarling: Ah yes, looks like newlines don't get handled nicely here — I think that's going to be an issue with `tfplugindocs` I think.

samstarling: Will see if I can improve this separately from this PR!

PR #123, opened 2025-02-06
Rely only on the OpenAPIv3 schema

isaacseymour:
We've deprecated our OpenAPIv2 schema, in favour of OpenAPIv3.

We were already using this for the generated client, but not for the docstrings.

I've also updated the instructions for copying the schema across from source and synced across the latest schema.


DEVELOPING.md
```diff
@@ -23,23 +23,11 @@ export INCIDENT_API_KEY="inc_development_<token>"
...
-If you need to regenerate the client, you first need to copy the following files over from `core/server/api/public/gen/http/`:
-
-```
-openapi.json
-openapi.yaml
-openapi3-secret.json
-openapi3.json
-openapi3.yaml
-```
+If you need to regenerate the client, you first need to copy the following files from `core`:
```
henryjcee, line 26: missing a file in here?

docs/resources/escalation_path.md
henryjcee, line 1: what is the cause of this diff and why is it only affecting escalation paths?

DEVELOPING.md
isaacseymour: no I just forgot to change the pluralisation. It's just one file now!

henryjcee (approved):

docs/resources/escalation_path.md
isaacseymour: Ah, it's because the v3 schema is missing some attribute descriptions. Big sad. Let me see if I can fix it.

PR #124, opened 2025-02-06
Make timezone a RequiresReplace field in the schedule resource

henryjcee:
Grab bag of changes in here that I fixed as I found them:

- **The main event:** Modifying the timezone of a schedule resource will now require replacement instead of planning correctly and failing on apply. 
- Fixed an acceptance test that was breaking because we were missing an s.
- Added a validator for rotation working intervals that will now prevent an empty list being provided. This would previously have planned fine and broken on the apply so the impact on anyone who's currently got this in config should be minimal.
- Improved error messages for custom fields and severities when they failed to apply.
- These changes have been tested by creating resources and simulating a provider upgrade. I have pretty good confidence that these don't break backwards compat
Also fixed a bug with schedule integration tests where we were failing to handle working hours correctly.

Screenshot 2025-02-10 at 10 39 24

### DIFF
--- a/internal/provider/incident_schedule_resource_test.go
+++ b/internal/provider/incident_schedule_resource_test.go
@@ -1,6 +1,7 @@
 package provider
 
 import (
+	"fmt"
 	"reflect"
 	"strconv"
 	"testing"
@@ -58,13 +59,6 @@ func TestAccIncidentScheduleResource(t *testing.T) {
 										Name: lo.ToPtr("Primary Layer One"),
 									},
 								},
-								WorkingInterval: &[]client.ScheduleRotationWorkingIntervalV2{
-									{
-										StartTime: "09:00",
-										EndTime:   "17:00",
-										Weekday:   "monday",
-									},
-								},
 							},
 						},
 					},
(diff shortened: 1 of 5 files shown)


internal/provider/incident_schedule_resource_test.go
```diff
@@ -94,6 +88,221 @@ func TestAccIncidentScheduleResource(t *testing.T) {
...
+			},
+			// Remove holiday config
+			{
+				Config: testAccIncidentScheduleResourceConfig(&client.ScheduleV2{
+					Name:     "holiday-test",
+					Timezone: "Europe/London",
+				}),
+				Check: resource.TestCheckResourceAttr(
+					"incident_schedule.example", "holidays_public_config.#", "0",
+				),
```
henryjcee, line 300: Doesn't delete the holiday config, just reduces the list to zero.

internal/provider/incident_schedule_resource_test.go
```diff
@@ -192,7 +400,7 @@ func generateVersionsArray(versions []client.ScheduleRotationV2) string {
 		result += "    layers          = " + generateLayersArray(version.Layers) + "\n"
 		result += "    users           = " + generateUsersArray(version.Users) + "\n"
 		if version.WorkingInterval != nil {
-			result += "    working_interval = " + generateWorkingIntervalsArray(version.WorkingInterval) + "\n"
+			result += "    working_intervals = " + generateWorkingIntervalsArray(version.WorkingInterval) + "\n"
```
henryjcee, line 403: :rip

isaacseymour (approved):

PR #125, opened 2025-02-10
Add alert attribute resource

henryjcee:


henryjcee:
Gonna add a service description for alert attributes so that it gets pulled into here.

internal/provider/incident_alert_attribute_resource.go
```diff
@@ -0,0 +1,192 @@
...
+				},
+			},
+			"name": schema.StringAttribute{
+				MarkdownDescription: apischema.Docstring("AlertAttributeV2", "name"),
+				Required:            true,
+			},
+			"type": schema.StringAttribute{
+				MarkdownDescription: apischema.Docstring("AlertAttributeV2", "type"),
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
```
isaacseymour, line 62: I don't think this is true by the way!

internal/provider/incident_alert_attribute_resource_test.go
```diff
@@ -0,0 +1,97 @@
...
+				),
+			},
+			// Block use of `Priority` as an attribute name
+			{
+				Config: testAccAlertAttributeResourceConfig(alertAttributeElement{
+					Name:  "Priority",
+					Type:  "String",
+					Array: false,
+				}),
+				ExpectError: regexp.MustCompile("cannot have an attribute named 'Priority'"),
```
isaacseymour, line 59: this is coming from the API right? (that's fine, just confused for a second!)

isaacseymour (approved):

internal/provider/incident_alert_attribute_resource_test.go
henryjcee: yeah, I'm not checking for the error at plan stage. We could do that but we rely on API validation basically everywhere else. Will stick a comment on this tho.

PR #126, opened 2025-02-10
Add alert attribute data source

henryjcee:


internal/provider/incident_alert_attribute_data_source_test.go
```diff
@@ -0,0 +1,77 @@
...
+  array = {{ .Array }}
+}
+data "incident_alert_attribute" "by_name" {
+  name         = incident_alert_attribute.example.name
+}
+`))
+
+func alertAttributeDefault() client.AlertAttributeV2 {
+	return client.AlertAttributeV2{
+		Name:  "Communications Lead",
```
isaacseymour, line 58: This is a slightly odd example 😂

isaacseymour (approved):

internal/provider/incident_alert_attribute_data_source_test.go
henryjcee: very fair

PR #127, opened 2025-02-10
ONC-5367 Add alert source resource

isaacseymour:
The first commit here moves all the engine stuff (param bindings, conditions, expressions) inside the `models` package, and standardises on a `.FromAPI(clientType)` and `.ToPayload()` interface for parsing responses and generating request bodies.

The second commit then uses that to add alert sources 🎉 

The third commit then adds a resource-level validation which prevents you from trying to use `jira_options` with a non-Jira-type source.


henryjcee (approved):
Have stared at this trying to find a problem for long enough. Looks sensible and I strongly approve of the new naming convention for the engine deser methods.

PR #87, opened 2024-06-27
chore(deps): upgrade to latest version of `oapi-codegen`

jamietanna:
Thanks for using the project :rocket:

PR #132, opened 2025-02-10
Update oapi-codegen

isaacseymour:
- Migrate from `deepmap/` to `oapi-codegen/oapi-codegen` (it moved)
- Bump the version
- Run it from `go generate` rather than `make`, for consistency

Closes #87


henryjcee (approved):

henryjcee (approved):
makes sense as discussed

PR #142, opened 2025-02-13
Alert source: mark email and secret token as non-changing

isaacseymour:
Once an alert source has been created, these won't change, so we need the 'use state for unknown' plan modifier, to ensure we don't get a diff on these.


henryjcee (approved):
Not totally clear to me why we need this. I think these'll now be marked as "known after apply" but I wouldn't expect a diff if these never change over the lifetime of the resource?

isaacseymour:
This is for updates! Without this it assumes the value will be recomputed on every mutation

PR #150, opened 2025-02-27
Support schema-only catalog attributes

isaacseymour:
Bunch of commits here:
- update the OpenAPIv3 schema and regen the client
- [BREAKING] Make `source_repo_url` required on `incident_catalog_type`: this means that all types managed in Terraform will get locked in the UI, and minimises the chance of state drift
- Add `schema_only` to `incident_catalog_attribute`: this allows attributes to be managed in Terraform, but leaves their _values_ to be managed in the incident.io dashboard.
- Add `managed_attributes` to `incident_catalog_entries` and `incident_catalog_entry`: this allows schema-only attributes' values to be left unchanged when `terraform apply` is run

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 ## Unreleased
 
+- Add `schema_only` as an option on `incident_catalog_type_attribute`. When set
+  to `true`, the attribute will be created from Terraform, but values for it will
+  be edited in the incident.io dashboard.
+- Add `managed_attributes` to both `incident_catalog_entry` and `incident_catalog_entries`. This
+  allows you to manage only some attributes of a catalog entry, while leaving others
+  unchanged. When attributes on a type are marked as `schema_only`, this avoids
+  Terraform trying to manage their state, allowing values set in the incident.io
+  dashboard to be preserved, and avoiding unnecessary diffs.
+
 ## v4.3.3
 
 - Fixed a panic when using an `incident_escalation_path` with a `notify_channel`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,11 @@
   to `true`, the attribute will be created from Terraform, but values for it must
   be edited in the incident.io dashboard.
 - Add `managed_attributes` to both `incident_catalog_entry` and `incident_catalog_entries`. This
-  allows you to manage only some attributes of a catalog entry, while leaving others
-  unchanged. When attributes on a type are marked as `schema_only`, this avoids
-  Terraform trying to manage their state, allowing values set in the incident.io
-  dashboard to be preserved, and avoiding unnecessary diffs.
+  allows you to manage only some attributes of a catalog entry, while leaving
+  others unchanged. This is most useful when attributes have been set as
+  `schema_only`, since it allows you to avoid Terraform trying to manage their
+  state, allowing values set in the incident.io dashboard to be preserved, and
+  avoiding unnecessary diffs.
 
 ## v4.3.3
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,11 @@
 ## Unreleased
 
+- BREAKING: `incident_catalog_type`'s `source_repo_url` attribute is now
+  required. This prevents the catalog type from being edited manually, and ensures
+  there is a link from the incident.io dashboard to the configuration that defines
+  the catalog type.
 - Add `schema_only` as an option on `incident_catalog_type_attribute`. When set
-  to `true`, the attribute will be created from Terraform, but values for it will
+  to `true`, the attribute will be created from Terraform, but values for it must
   be edited in the incident.io dashboard.
 - Add `managed_attributes` to both `incident_catalog_entry` and `incident_catalog_entries`. This
   allows you to manage only some attributes of a catalog entry, while leaving others
--- a/docs/resources/workflow.md
+++ b/docs/resources/workflow.md
@@ -85,11 +85,11 @@ resource "incident_workflow" "autoassign_incident_lead" {
 - `continue_on_step_error` (Boolean) Whether to continue executing the workflow if a step fails
 - `expressions` (Attributes Set) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
 - `include_private_incidents` (Boolean) Whether to include private incidents
-- `name` (String) The human-readable name of the workflow
+- `name` (String) Name provided by the user when creating the workflow
 - `once_for` (List of String) This workflow will run 'once for' a list of references
 - `runs_on_incident_modes` (List of String) Incidents in these modes will be affected by the workflow
-- `runs_on_incidents` (String) Which incidents should the workflow be applied to? (newly_created or newly_created_and_active)
-- `state` (String) The state of the workflow (e.g. is it draft, or disabled)
+- `runs_on_incidents` (String) Which incidents should the workflow be applied to?
+- `state` (String) What state this workflow is in
 - `steps` (Attributes List) Steps that are executed as part of the workflow (see [below for nested schema](#nestedatt--steps))
 - `trigger` (String) Unique name of the trigger
 
@@ -97,7 +97,7 @@ resource "incident_workflow" "autoassign_incident_lead" {
 
 - `delay` (Attributes) Configuration controlling workflow delay behaviour (see [below for nested schema](#nestedatt--delay))
 - `folder` (String) Folder to display the workflow in
-- `shortform` (String) Shortform used to trigger manual workflows in Slack - e.g. `/inc workflows page-ceo`
+- `shortform` (String) The shortform used to trigger this workflow (only applicable for manual triggers)
 
 ### Read-Only
 
--- a/internal/provider/incident_catalog_entries_resource.go
+++ b/internal/provider/incident_catalog_entries_resource.go
@@ -329,7 +329,7 @@ func (r *IncidentCatalogEntriesResource) buildModel(catalogType client.CatalogTy
 
 type catalogEntryModelPayload struct {
 	CatalogEntryID *string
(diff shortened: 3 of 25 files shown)


internal/provider/incident_catalog_entries_resource.go
```diff
@@ -600,3 +622,58 @@ func (r *IncidentCatalogEntriesResource) reconcile(ctx context.Context, data *In
...
+			// Weird but ok
+			return nil, false
+		}
+
+		managedAttrSet[attrIDStr.ValueString()] = true
+	}
+
+	// Technically this isn't race-safe, since multiple goroutines _could_ try to set this at once.
+	// However, it's a static value that will be the same however it's built, so not worried about this right now.
+	m.managedAttrSet = managedAttrSet
```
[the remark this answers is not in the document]
leonore: Lol, not copilot calling this out in spite of your comment!!

CHANGELOG.md
```diff
@@ -1,5 +1,18 @@
...
+- BREAKING: `incident_catalog_type`'s `source_repo_url` attribute is now
+  required. This prevents the catalog type from being edited manually, and ensures
+  there is a link from the incident.io dashboard to the configuration that defines
+  the catalog type.
+- Add `schema_only` as an option on `incident_catalog_type_attribute`. When set
+  to `true`, the attribute will be created from Terraform, but values for it must
+  be edited in the incident.io dashboard.
+- Add `managed_attributes` to both `incident_catalog_entry` and `incident_catalog_entries`. This
+  allows you to manage only some attributes of a catalog entry, while leaving others
+  unchanged. When attributes on a type are marked as `schema_only`, this avoids
```
leonore, line 12: This is a bit unclear to me. Do they _need_ to use schema_only on the attribute to be able to use this?

internal/provider/incident_catalog_entries_resource.go
```diff
@@ -600,3 +622,58 @@ func (r *IncidentCatalogEntriesResource) reconcile(ctx context.Context, data *In
 
 	return catalogType, entries, nil
 }
+
+// isAttributeManaged checks if the given attribute should be managed by this resource.
+func (m *IncidentCatalogEntriesResourceModel) isAttributeManaged(attributeID string) bool {
+	// Check if the attribute is in the managed list
+	attrSet, known := m.managedAttributesSet()
+	if !known {
+		return true
```
leonore, line 631: I think this comment is wrong - what this checks is that you have the list right. So if you don't have the list, all attributes are managed. Caught me out so I would update!

internal/provider/incident_catalog_entries_resource.go
```diff
@@ -600,3 +622,58 @@ func (r *IncidentCatalogEntriesResource) reconcile(ctx context.Context, data *In
...
+
+	return attrSet[attributeID]
+}
+
+func (m *IncidentCatalogEntriesResourceModel) managedAttributesSet() (map[string]bool, bool) {
+	if m.ManagedAttributes.IsNull() || m.ManagedAttributes.IsUnknown() {
+		return nil, false
+	}
+
+	if m.managedAttrSet != nil {
```
leonore, line 642: I think we should use len(m.managedAttrSet) == 0 here. If someone sets it empty?

docs/resources/workflow.md
```diff
@@ -85,19 +85,19 @@ resource "incident_workflow" "autoassign_incident_lead" {
 - `continue_on_step_error` (Boolean) Whether to continue executing the workflow if a step fails
 - `expressions` (Attributes Set) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
 - `include_private_incidents` (Boolean) Whether to include private incidents
-- `name` (String) The human-readable name of the workflow
+- `name` (String) Name provided by the user when creating the workflow
```
leonore, line 88: This is less clear imo lol

internal/provider/incident_catalog_entries_resource.go
```diff
@@ -600,3 +622,58 @@ func (r *IncidentCatalogEntriesResource) reconcile(ctx context.Context, data *In
...
+		return m.managedAttrSet, true
+	}
+
+	managedAttrSet := map[string]bool{}
+	for _, attrElem := range m.ManagedAttributes.Elements() {
+		if attrElem.IsUnknown() {
```
leonore, line 648: I'm unfamiliar with `types.Set` - if just the attrElem is unknown/not a string, why are we exiting the whole loop?

internal/provider/incident_catalog_entries_resource.go
```diff
@@ -600,3 +622,58 @@ func (r *IncidentCatalogEntriesResource) reconcile(ctx context.Context, data *In
...
+	}
+
+	// Technically this isn't race-safe, since multiple goroutines _could_ try to set this at once.
+	// However, it's a static value that will be the same however it's built, so not worried about this right now.
+	m.managedAttrSet = managedAttrSet
+	return managedAttrSet, true
+}
+
+func (m *IncidentCatalogEntriesResourceModel) buildUpdateAttributes() *[]string {
+	if m.ManagedAttributes.IsNull() {
```
leonore, line 668: || m.ManagedAttributes.IsUnknown()? should we have a helper for this, since we use it twice?

internal/provider/incident_catalog_entry_resource.go
```diff
@@ -259,12 +280,24 @@ func (r *IncidentCatalogEntryResource) Update(ctx context.Context, req resource.
...
 	}
 
-	result, err := r.client.CatalogV2UpdateEntryWithResponse(ctx, data.ID.ValueString(), client.UpdateEntryRequestBody{
-		Name:            data.Name.ValueString(),
-		Rank:            rank,
-		ExternalId:      data.ExternalID.ValueStringPointer(),
-		Aliases:         &aliases,
-		AttributeValues: data.buildAttributeValues(),
+	var updateAttributes *[]string
+	if !data.ManagedAttributes.IsUnknown() && !data.ManagedAttributes.IsNull() {
```
leonore, line 284: see comment above!

internal/provider/incident_catalog_entry_resource.go
```diff
@@ -259,12 +280,24 @@ func (r *IncidentCatalogEntryResource) Update(ctx context.Context, req resource.
...
+		var attributeIDs []string
+		diags := data.ManagedAttributes.ElementsAs(ctx, &attributeIDs, false)
```
leonore, line 286: actually you have a function for this, no?

internal/provider/incident_catalog_entry_resource.go
```diff
@@ -296,9 +329,88 @@ func (r *IncidentCatalogEntryResource) ImportState(ctx context.Context, req reso
...
+			// doesn't seem to expose this, so we just have to skip the check.
+			continue
+		}
+
+		attributeID := attributeValue.Attribute.ValueString()
+		if !managedAttributesMap[attributeID] {
+			resp.Diagnostics.AddAttributeError(
+				path.Root("attribute_values").AtListIndex(idx),
+				"Unmanaged Attribute",
+				fmt.Sprintf("Attribute ID %q is specified in attribute_values but is not in the managed_attributes set. "+
```
leonore, line 372: Should we not only say this for schema_only attributes?

leonore (approved):
just some confusions around validation/helpers, but otherwise i think makes sense!

internal/provider/incident_catalog_entry_resource.go
isaacseymour: In this resource, we don't actually know whether the attribute is schema-only or not: we just have its ID :(

docs/resources/workflow.md
isaacseymour: This comes from the OpenAPI schema fyi!

internal/provider/incident_catalog_entries_resource.go
isaacseymour: imo if someone sets it to `[]` that means "no attributes are managed" vs not setting it which means "all attributes are managed".

It wouldn't be _totally_ wild to only manage the `name` / `external_id` / `aliases` in tf and have a bunch of dashboard-managed attributes imo!

internal/provider/incident_catalog_entries_resource.go
isaacseymour: Yeah this took me literally 4 hours to figure out - worst bit of tf I've encountered so far.

An unknown value is (e.g.) a reference to another resource, but one that doesn't exist yet. So if I do:
```hcl
resource "incident_catalog_type" "a" {
  ...
}

resource "incident_catalog_type_attribute" "a_b" {
  catalog_type_id = incident_catalog_type.a.id
}
```
... then the value of `CatalogTypeID` in the attribute resource will be unknown at plan-time, since we've not created either resource _yet_.

When it comes to set / list types, the whole set/list could be unknown (if it uses a reference), or alternatively the list/set might be known but have some unknown elements, e.g.:
```hcl
  managed_attributes = [
    incident_catalog_type_attribute.already_exists.id,
    incident_catalog_type_attribute.new_one.id,
  ]
```

In that case:
- the set is known
- the first element is known (since it already exists)
- the second element is an unknown string

I decided that the easiest approach here is to just give up if anything in the set is unknown, since it seems really confusing to me to try to _partially_ apply managed attributes, and I couldn't figure out a way of getting the reference that the unknown thing points at, since comparing that would also be useful.

Does that make sense? (I will comment)

internal/provider/incident_catalog_entries_resource.go
isaacseymour: Same as the comment above!

internal/provider/incident_catalog_entry_resource.go
isaacseymour: `ElementsAs` returns an error if any of the values are unknown which is actually very convenient here - if we try to call this API and some values are unknown that would be very freaky and we should abort!

PR #161, opened 2025-03-21
Team ID support in schedules and escalation paths again

pipt:
But using `UseStateForUnknown` on `team_ids` so that we don't break when the terraform state is from a previous version of the provider.


internal/provider/incident_escalation_path_resource.go
```diff
@@ -127,6 +128,9 @@ func (r *IncidentEscalationPathResource) Schema(ctx context.Context, req resourc
 				Optional:            true,
 				ElementType:         types.StringType,
 				Computed:            true,
+				PlanModifiers: []planmodifier.List{
```
isaacseymour, line 131: Could we use `SetAttribute` / `planmodifier.Set` etc here? I don't think we care about the order of these IDs, so Set is better than List!

PR #162, opened 2025-03-24
Add support for setting team IDs on escalation paths and schedules

pipt:
This is the same as #159 with the following changes:
- Using a set attribute for `team_ids` rather than a list attribute because we don't care about the order
- Setting a default for `team_ids` so we don't error when team IDs isn't already in the state

### DIFF
--- a/docs/resources/escalation_path.md
+++ b/docs/resources/escalation_path.md
@@ -116,6 +116,9 @@ resource "incident_escalation_path" "urgent_support" {
       ]
     }
   ]
+
+  # Teams that use this escalation path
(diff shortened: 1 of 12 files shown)


docs/resources/escalation_path.md
```diff
@@ -129,6 +132,7 @@ resource "incident_escalation_path" "urgent_support" {
 
 ### Optional
 
+- `team_ids` (Set of String) IDs of teams that own this schedule
```
leonore, line 135: Doesn't this mean that it will generate a different state every time?

pipt: No, when it compares the current state to the remote state, it'll do so without caring about the order. That will prevent us incorrectly saying that there's a difference if the API returns the team IDs in a different order to what's in the state.

leonore (approved):

PR #173, opened 2025-04-09
Implement import for catalog_type_attribute resource

isaacseymour:
I've also added `terraform import` syntax to all the resources that implement it

### DIFF
--- a/examples/resources/incident_status/import.sh
+++ b/examples/resources/incident_status/import.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+# Import a status using its ID
+# Replace the ID with a real ID from your incident.io organization
+terraform import incident_status.example 01ABC123DEF456GHI789JKL
--- a/internal/provider/incident_catalog_type_attribute_resource.go
+++ b/internal/provider/incident_catalog_type_attribute_resource.go
@@ -3,10 +3,12 @@ package provider
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync"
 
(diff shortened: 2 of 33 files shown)


examples/resources/incident_status/import.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
leonore, line 1: Did we already support importing for all these other resources?

leonore (approved):

examples/resources/incident_status/import.sh
isaacseymour: Yes! Just hadn't put it in the docs 😭

internal/provider/incident_catalog_type_attribute_resource.go
```diff
@@ -441,6 +444,27 @@ func (r *IncidentCatalogTypeAttributeResource) lockFor(ctx context.Context, cata
 	return do(ctx, typeResult.JSON200.CatalogType)
 }
 
+func (r *IncidentCatalogTypeAttributeResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// The import ID format is catalogTypeID:attributeID
+	idParts := strings.Split(req.ID, ":")
```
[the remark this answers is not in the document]
isaacseymour: https://developer.hashicorp.com/terraform/plugin/framework/resources/import#multiple-attributes

Taken directly from here 😭

PR #190, opened 2025-04-28
Change custom_fields to be a set

shlokcshah55:
Terraform plans were was giving a diff when the order or the custom fields were not matching the database because the custom fields were stored as a list, changing this to a set means regardless of the order of the attributes it wont show a diff.

### DIFF
--- a/internal/provider/incident_alert_route_resource_test.go
+++ b/internal/provider/incident_alert_route_resource_test.go
@@ -352,7 +352,7 @@ func testAccIncidentAlertRouteWithAlphabeticalCustomFields(name string) string {
               literal = "First alphabetical custom field value"                                                                                                                       
             }                                                                                                                                                                          
           }                                                                                                                                                                            
-        ],                                                                                                                                                                            
+        },                                                                                                                                                                            
         {                                                                                                                                                                          
           custom_field_id = incident_custom_field.alpha_field2.id                                                                                                                      
           merge_strategy = "first-wins"                                                                                                                                                
--- a/internal/provider/incident_alert_route_resource_test.go
+++ b/internal/provider/incident_alert_route_resource_test.go
@@ -218,7 +218,7 @@ func TestAccIncidentCustomFieldsAlphabeticalOrder(t *testing.T) {
 		Steps: []resource.TestStep{
 			// First step - create alert route with custom fields in alphabetical order
 			{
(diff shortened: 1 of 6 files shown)


internal/provider/incident_alert_route_resource_test.go
```diff
@@ -211,6 +211,206 @@ func TestAccIncidentAlertRouteResourceConditionalConfig(t *testing.T) {
...
+    }                                                                                                                                                                                  
+                                                                                                                                                                                       
+    incident_template = {                                                                                                                                                              
+      # Focus on custom fields in alphabetical order                                                                                                                                   
+      custom_fields = [                                                                                                                                                                
+       {                                                                                                                                                                          
+          custom_field_id = incident_custom_field.alpha_field1.id
```
phroggyy, line 348: I don't think this'll test what you want it to - these are correctly ordered to begin with, so the change to a set makes no difference. You'll want to specify these in the opposite order, such that the field created last is set first

internal/provider/incident_alert_route_resource_test.go
```diff
@@ -211,6 +211,206 @@ func TestAccIncidentAlertRouteResourceConditionalConfig(t *testing.T) {
...
+    field_type  = "text"                                                                                                                                                               
+  }                                                                                                                                                                                    
+                                                                                                                                                                                       
+  resource "incident_custom_field" "alpha_field2" {                                                                                                                                    
+    name        = "Alpha Test Field 2"                                                                                                                                                 
+    description = "Second alphabetical custom field"                                                                                                                                   
+    field_type  = "text"
```
phroggyy, line 262:
```suggestion
    field_type  = "text"
    depends_on = [incident_custom_field.alpha_field_1]
```

This'll ensure field 2 is always created after field 1

internal/provider/incident_alert_route_resource_test.go
```diff
@@ -211,6 +211,208 @@ func TestAccIncidentAlertRouteResourceConditionalConfig(t *testing.T) {
...
+      condition_groups     = []                                                                                                                                                        
+      defer_time_seconds   = 300                                                                                                                                                       
+      grouping_keys        = []                                                                                                                                                        
+    }                                                                                                                                                                                  
+                                                                                                                                                                                       
+    incident_template = {                                                                                                                                                              
+      # Focus on custom fields in alphabetical order
```
phroggyy, line 347:
```suggestion
      # Focus on custom fields in reverse lexicographical order                                                                                                                                   
```

PR #201, opened 2025-05-14
Clarify team_ids documentation

leonore:
We updated this documentation in our API, but it was not pulled through properly.


internal/apischema/public-schema-v3-including-secret-endpoints.json
isaacseymour, line 1: we should mark this file as generated (in `.gitattributes`)!

PR #210, opened 2025-05-27
Use `Set` to model schedule rotations

### LINKED ISSUE #202: Use a set for schedule rotations in terraform, not a list
When exporting a complex schedule, the UI exports rotations in the schedule in a different order to how your DB stores them. This leads to very large terraform plan diffs when looking at larger schedules.

I also noted that sometimes the export functionality would output the days of the week in a different order to what is stored in the DB leading to diffs here too.

I am unsure if this is a bug in the provider itself or an issue with the export feature in the UI.

Steps to reproduce:

1. Take a large complex schedule
2. Use the "Copy this schedule to Terraform" feature in incident
3. Import the resource to TF state
4. Run `terraform plan`
5. The `rotations` array will likely be out of order showing quite an alarming diff

edds:
The order isn't important, using a set will allow us to return them in a different order as they are defined without causing a large plan diff.

Fixes #202

### DIFF
--- a/internal/provider/incident_schedule_resource.go
+++ b/internal/provider/incident_schedule_resource.go
@@ -76,7 +76,7 @@ func (r *IncidentScheduleResource) Schema(ctx context.Context, req resource.Sche
 				},
 				Optional: true,
 			},
-			"rotations": schema.ListNestedAttribute{
+			"rotations": schema.SetNestedAttribute{
 				NestedObject: schema.NestedAttributeObject{
 					Attributes: map[string]schema.Attribute{
 						"id": schema.StringAttribute{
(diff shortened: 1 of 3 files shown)


internal/provider/incident_schedule_resource.go
```diff
@@ -76,7 +76,7 @@ func (r *IncidentScheduleResource) Schema(ctx context.Context, req resource.Sche
 				},
 				Optional: true,
 			},
-			"rotations": schema.ListNestedAttribute{
+			"rotations": schema.SetNestedAttribute{
 				NestedObject: schema.NestedAttributeObject{
 					Attributes: map[string]schema.Attribute{
 						"id": schema.StringAttribute{
```
leonore, line 82: I think you'll need the `versions` below to be a set as well!

rorymalcolm:
qq: Will we need a plan modifier here if we're changing the underlying type? I'm unsure

internal/provider/incident_schedule_resource.go
edds: Good shout, I've made that change also :+1:

rorymalcolm:
> @rorymalcolm my understanding (and from testing locally) is no. In the terraform state lists and sets are stored as the same data structure, so it should cleanly map over for any consumers.

Ah nice one, thanks for checking!

PR #229, opened 2025-06-23
Make schedule handovers required

phroggyy:
Our API requires these already, so the previous terraform schema was incorrect, leading to plans saying this would succeed, for it then to get a 422 error response back.


leonore (commented):
should we change our public API as well? I don't think it's a breaking change given we've always validated this

leonore (approved):
you'll need to fix tests! and please can you add a note in the unreleased changelog bit?

PR #230, opened 2025-06-24
Consistently sort keys for JSON values returned as literal engine values

estephinson:
In some places (such as initially setting up an alert source), the engine literal values for complex objects like templatedtext will be stored un-sorted.

Functionally this doesn't impact behaviour, but it does cause issues with terraform state, due to them being string values, and terraform produces alphabetically sorted JSON objects from `jsonencode`.

### DIFF
--- a/internal/provider/models/engine_test.go
+++ b/internal/provider/models/engine_test.go
@@ -1,0 +1,57 @@
+package models
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/incident-io/terraform-provider-incident/internal/client"
+)
+
+func TestIncidentEngineParamBindingValue_JSONOrdering(t *testing.T) {
+	tests := []struct {
+		name               string
+		apiJSON            string
+		expectedNormalized string
+		description        string
+	}{
+		{
+			name:               "keys_should_be_sorted_lexicographically",
+			apiJSON:            `{"type":"doc","content":[{"type":"paragraph","content":[{"type":"varSpec","attrs":{"name":"description","label":"Payload → Description","missing":false}}]}]}`,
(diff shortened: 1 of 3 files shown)


internal/provider/models/engine_test.go
```diff
@@ -0,0 +1,57 @@
...
+			name:               "already_sorted_should_remain_unchanged",
+			apiJSON:            `{"content":[{"content":[{"attrs":{"label":"Payload → Description","missing":false,"name":"description"},"type":"varSpec"}],"type":"paragraph"}],"type":"doc"}`,
+			expectedNormalized: `{"content":[{"content":[{"attrs":{"label":"Payload → Description","missing":false,"name":"description"},"type":"varSpec"}],"type":"paragraph"}],"type":"doc"}`,
+			description:        "JSON with lexicographically sorted keys should remain unchanged",
+		},
+		{
+			name:               "nested_objects_keys_sorted_lexicographically",
+			apiJSON:            `{"type":"doc","content":[{"type":"paragraph","content":[{"attrs":{"name":"description","missing":false,"label":"Payload → Description"},"type":"varSpec"}]}]}`,
+			expectedNormalized: `{"content":[{"content":[{"attrs":{"label":"Payload → Description","missing":false,"name":"description"},"type":"varSpec"}],"type":"paragraph"}],"type":"doc"}`,
+			description:        "All nested object keys should be sorted lexicographically",
```
leonore, line 34: can we add a test for content that's just a simple string, just in case?

leonore (approved):
I would test with an alert source locally to make sure you no longer get the provider diff errors - and can we add the change under unreleased in the changelog to make traceability easier?

estephinson:
> and can we add the change under unreleased in the changelog to make traceability easier?

have done

> I would test with an alert source locally to make sure you no longer get the provider diff errors

I actually haven't been able to reproduce it with some terraform setup yet, despite my best efforts - will try again shortyl

PR #236, opened 2025-07-18
Add incident_catalog_entries data source

leonore:
This adds an incident_catalog_entries data source, which means you can manage your catalog via catalog-importer but then refer to them via terraform, if you want.

### DIFF
--- a/internal/provider/incident_catalog_entries_data_source.go
+++ b/internal/provider/incident_catalog_entries_data_source.go
@@ -1,0 +1,187 @@
+package provider
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/datasource"
+	"github.com/hashicorp/terraform-plugin-framework/datasource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	"github.com/incident-io/terraform-provider-incident/internal/apischema"
+	"github.com/incident-io/terraform-provider-incident/internal/client"
+)
+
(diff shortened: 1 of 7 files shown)


internal/provider/incident_catalog_entries_data_source.go
```diff
@@ -0,0 +1,187 @@
...
+	return &IncidentCatalogEntriesDataSource{}
+}
+
+type IncidentCatalogEntriesDataSource struct {
+	client *client.ClientWithResponses
+}
+
+type IncidentCatalogEntriesDataSourceModel struct {
+	CatalogTypeID  types.String                          `tfsdk:"catalog_type_id"`
+	CatalogEntries []IncidentCatalogEntryDataSourceModel `tfsdk:"catalog_entries"`
```
leonore, line 31: this reuses the incident_catalog_entry_data_source model but happy to be told this is the wrong approach?

internal/provider/incident_catalog_entries_data_source.go
```diff
@@ -0,0 +1,187 @@
...
+	// Fetch entries for the specified catalog type
+	catalogTypeID := data.CatalogTypeID.ValueString()
+	params := &client.CatalogV3ListEntriesParams{
+		CatalogTypeId: catalogTypeID,
+	}
+
+	// Paginate through all entries
+	params.PageSize = 100
+
+	for {
```
phroggyy, line 73: without any filtering support, I wonder if this might cause some sadness if you have like 40k entries on your type 🤔 I think it's _probably_ okay though

phroggyy (approved):

PR #243, opened 2025-08-04
Support ack_mode on escalation path level nodes

estephinson:
This adds support to the provider to supply an `ack_mode` to escalation path node levels.

This will default to `all` on the backend, so completely fine to not supply. Options are `first` and `all`.

### DIFF
--- a/internal/provider/incident_escalation_path_resource.go
+++ b/internal/provider/incident_escalation_path_resource.go
@@ -219,6 +219,7 @@ func (r *IncidentEscalationPathResource) getPathSchema(depth int) schema.NestedA
 						MarkdownDescription: apischema.Docstring(
 							"EscalationPathNodeLevelV2", "ack_mode"),
 						Optional: true,
+						Computed: true,
 						Default:  stringdefault.StaticString("all"),
 					},
 				},


docs/resources/escalation_path.md
```diff
@@ -452,6 +452,7 @@ Required:
 
 Optional:
 
+- `ack_mode` (String) Controls the behaviour of acknowledgements for this level, with 'first' cancelling all other escalations on the same level when someone acks
```
isaacseymour, line 455: This only explains 'first' mode - what are the other modes?

internal/provider/incident_escalation_path_resource.go
```diff
@@ -212,6 +214,11 @@ func (r *IncidentEscalationPathResource) getPathSchema(depth int) schema.NestedA
 							"EscalationPathNodeLevelV2", "time_to_ack_weekday_interval_config_id"),
 						Optional: true,
 					},
+					"ack_mode": schema.StringAttribute{
+						MarkdownDescription: apischema.Docstring(
+							"EscalationPathNodeLevelV2", "ack_mode"),
+						Optional: true,
```
isaacseymour, line 220: Should this have a default defined as well..?

isaacseymour (approved):

PR #246, opened 2025-08-12
Add use_name_as_identifier field to catalog types

rorydbain:
Adds the missing `use_name_as_identifier` field to the `incident_catalog_type` resource.

This field was available in the API but missing from the Terraform provider. Now users can set it to automatically include the catalog entry name in the list of identifiers.

- Added `use_name_as_identifier` field (optional, defaults to false)
- Updated API schema and regenerated client
- Added tests for create/update scenarios
- Updated example docs

Also fixes some issues where custom field and alert route acceptance tests were failing due to naming collisions with existing resources.


internal/provider/incident_alert_route_resource_test.go
rorydbain, line 1: These unrelated tests were failing because of naming collisions

PR #249, opened 2025-08-21
feat: Add annotations support to IncidentSchedule resource

maxtacu:
Incident io API supports annotations for schedules v2. Terraform provider does not allow setting custom annotations although it uses them to inject terraform version used. 
This PR adds feature to add custom annotations for schedules in incident-io.


isaacseymour:
Hey @maxtacu - can I ask what annotations you'd like to manage with Terraform?

Annotations are intentionally not included in this provider, since the provider uses them itself to mark the schedule as managed in code, and other annotations don't have any effect on the schedule!

PR #250, opened 2025-08-22
ONC-7504 Fix normaliseJSON for HTML-ish characters

isaacseymour:
Setting `SetEscapeHTML(false)` means using HTML characters like `>` somehow ends up getting screwed up: I think this is because Terraform's `jsonencode` _does_ escape the HTML, although I've had a dig through the implementation inside `cty` and can't figure out _why_ that happens.

Anyway, removing that has fixed this test case

### DIFF
--- a/internal/provider/models/engine.go
+++ b/internal/provider/models/engine.go
@@ -138,7 +138,6 @@ func normaliseJSON(jsonString string) (string, error) {
 	// Use encoder with HTML escaping disabled to preserve special characters like >
 	var buf bytes.Buffer
 	encoder := json.NewEncoder(&buf)
-	encoder.SetEscapeHTML(false)
 	err = encoder.Encode(data)
 	if err != nil {
 		return "", err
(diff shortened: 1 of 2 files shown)


internal/provider/models/engine.go
```diff
@@ -138,7 +138,6 @@ func normaliseJSON(jsonString string) (string, error) {
 	// Use encoder with HTML escaping disabled to preserve special characters like >
 	var buf bytes.Buffer
 	encoder := json.NewEncoder(&buf)
-	encoder.SetEscapeHTML(false)
```
rorydbain, line 141: It looks like Ed added this to fix a case where our response was varying due to decoding (https://github.com/incident-io/terraform-provider-incident/pull/240) - just checking you saw that prior change?

isaacseymour: I think the tests now cover both the cases there (`->` and `→`), so I think that change was just incorrect

rorydbain (approved):

PR #273, opened 2025-10-28
fix: error handling when retrieving resources from incident.io

maxtacu:
this is a continuation MR from https://github.com/incident-io/terraform-provider-incident/pull/257 to cover all remaining resources in incident-io provider

This PR fixes a critical error handling issue where `result.StatusCode()` was being called on API responses that don't actually return a `StatusCode()` method, causing runtime panics and incorrect error handling throughout the provider.


mbfisher (approved):
Amazing, thank you for this @maxtacu!! It's the end of the work day here and since there's quite a few changes I'm going to merge this tomorrow, just in case 🙏

PR #284, opened 2025-11-20
feat: Add incident_incident_types data source from incident.io API

maxtacu:
This new data source allows users to retrieve all configured incident types, including their names, descriptions, and settings like whether they are default or private-only. It follows the same pattern as other data sources in the provider.


samstarling (commented):
Hey @maxtacu! Thanks for the contribution, this is looking good. Just a couple of comments — one on error handling, and another on the inclusion of timestamps. Once those are addressed I'd be very happy to merge this.

samstarling:
Looks good, now that I understand the `result.StatusCode()` issue you'd flagged recently, thank you!

PR #299, opened 2025-12-23
Fix schema_only drift when attribute mode changes

vstpme:
When a catalog type attribute with `schema_only=true` is used as a team's escalation path, the system changes its mode from `dashboard` to `internal`. The provider only recognized `dashboard` as schema-only, causing it to compute `schema_only=false` on read and triggering Terraform drift.

This PR makes it so that `internal` is recognized as a valid schema-only mode. The same goes for `external` and `dynamic` modes.

<details>

<summary>Here's a Terraform file I used for testing</summary>

```tf

# STEP 1: Create custom team catalog type and attributes
resource "incident_catalog_type" "example_team_type" {
  name            = "Example Team Type"
  description     = "Custom team type for testing"
  source_repo_url = "https://example.com"
}

resource "incident_catalog_type_attribute" "example_team_type_members" {
  catalog_type_id = incident_catalog_type.example_team_type.id
  name            = "Members"
  type            = "User"
  array           = true
}

resource "incident_catalog_type_attribute" "example_team_type_escalation_path" {
  catalog_type_id = incident_catalog_type.example_team_type.id
  name            = "Escalation Path"
  type            = "EscalationPath"
  array           = true
  schema_only     = true
}

resource "incident_catalog_entry" "example_team" {
  catalog_type_id    = incident_catalog_type.example_team_type.id
  name               = "Example Team"
  external_id        = "example-team"
  managed_attributes = []
  attribute_values   = []
}

# STEP 2: Go to Teams > Configuration in the dashboard and set "Example Team Type" as the Team type

# STEP 3: Create a schedule and an escalation path referencing the team (uncomment after performing steps 1 and 2)
# resource "incident_schedule" "example_schedule" {
#   name     = "Example Schedule"
#   timezone = "Europe/London"
#
#   rotations = [{
#     id   = "primary"
#     name = "Primary"
#     versions = [{
#       handover_start_at = "2024-01-01T09:00:00Z"
#       users             = []
#       layers = [{
#         id   = "layer-1"
#         name = "Layer 1"
#       }]
#       handovers = [{
#         interval_type = "daily"
#         interval      = 1
#       }]
#     }]
#   }]
#
#   team_ids = [incident_catalog_entry.example_team.id]
# }
#
# resource "incident_escalation_path" "example_escalation_path" {
#   name = "Example Escalation Path"
#
#   path = [{
#     id   = "start"
#     type = "level"
#     level = {
#       targets = [{
#         type    = "schedule"
#         id      = incident_schedule.example_schedule.id
#         urgency = "high"
#       }]
#       time_to_ack_seconds = 300
#     }
#   }]
#
#   team_ids = [incident_catalog_entry.example_team.id]
# }

```

</details>

### DIFF
--- a/internal/provider/incident_catalog_type_attribute_resource_test.go
+++ b/internal/provider/incident_catalog_type_attribute_resource_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/hashicorp/terraform-plugin-testing/terraform"
+	"github.com/samber/lo"
 
 	"github.com/incident-io/terraform-provider-incident/internal/client"
 )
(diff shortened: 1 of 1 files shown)


internal/provider/incident_catalog_type_attribute_resource.go
```diff
@@ -30,6 +30,24 @@ var (
...
+// - external: values managed by external integrations
+// - dynamic: values dynamically computed.
+func isSchemaOnlyMode(mode client.CatalogTypeAttributeV3Mode) bool {
+	switch mode {
+	case client.CatalogTypeAttributeV3ModeDashboard,
+		client.CatalogTypeAttributeV3ModeInternal,
+		client.CatalogTypeAttributeV3ModeExternal,
+		client.CatalogTypeAttributeV3ModeDynamic:
+		return true
+	default:
```
edds: I'm not sure if it's worth being explicit here and having `client.CatalogTypeAttributeV3ModeAPI`.

internal/provider/incident_catalog_type_attribute_resource_test.go
```diff
@@ -136,3 +137,45 @@ func testAccIncidentCatalogTypeAttributeResourceConfig(attribute client.CatalogT
 
 	return buf.String()
 }
+
+func TestAttributeToPayload_PreservesModes(t *testing.T) {
```
vstpme: this is unfortunately the best test I could come up with. I tried to write a proper acceptance test, but couldn't find a good way of updating an attribute to make it internal through the public API that's consistent with how it works via the internal API when setting the team type in settings

PR #313, opened 2026-02-05
Add validation for `catalog_type_attribute` `type` field.

ab:
Validate that the type field only accepts valid primitive types (String, Text, Number, Bool) or Custom["TypeName"] references. This prevents users from using invalid types like DateTime or lowercase variations.

:warning: **Warning:** This is untested code written by Claude, and I have no idea whether it actually does what it claims to do. (Hence why I'm marking this as a draft pull request.) But maybe it's a useful starting point for someone else to take this on.

### DIFF
--- a/internal/provider/incident_catalog_type_attribute_resource.go
+++ b/internal/provider/incident_catalog_type_attribute_resource.go
@@ -15,6 +15,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/listplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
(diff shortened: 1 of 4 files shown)


benfdking:
Thanks for the starting point — I've dug into this against the API, and I don't think we should land client-side validation here, so closing.

The set of valid attribute types isn't static: the API builds it per-request from the org's own non-archived catalog types plus the catalog primitives. That means it legitimately includes bare non-`Custom` type names (`SlackChannel`, `PagerDutyService`, `Schedule`, `User`, and 100+ others), dynamic type names like `ServiceNowChoiceList["incident.priority"]`, and the `Image["avatar"]` primitive — all of which this validator rejects. Referencing another catalog type by its `type_name` is documented, common usage, so this would break existing valid configs at `plan` with no escape hatch. The API also skips the type check entirely for `mode = dynamic` attributes.

The API already returns a clear `unrecognised attribute type 'X'` validation error, so the upside would be moving that from apply to plan, but it is a little complicated as it could also depend on resources created in the plan itself.

PR #322, opened 2026-03-11
Add `repeat_config` support to escalation path resource

harshkheskani:
* Adds repeat_config block to the incident_escalation_path resource with repeat_after_seconds and delay_repeat_on_activity attributes
* Adds repeat_config to the escalation path data source (computed)
* Regenerates client from latest public API spec which includes the new EscalationPathRepeatConfigV2 type

### DIFF
--- a/docs/resources/escalation_path.md
+++ b/docs/resources/escalation_path.md
@@ -130,7 +130,7 @@ resource "incident_escalation_path" "urgent_support" {
 
 ### Optional
 
-- `repeat_config` (Attributes) (see [below for nested schema](#nestedatt--repeat_config))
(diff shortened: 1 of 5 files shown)


docs/resources/escalation_path.md
```diff
@@ -4002,6 +4003,15 @@ Required:
 
 
 
+<a id="nestedatt--repeat_config"></a>
+### Nested Schema for `repeat_config`
+
+Required:
+
+- `delay_repeat_on_activity` (Boolean)
```
leonore: Can't remember where we get this from, but can we add a tad bit more copy explaining this?

PR #329, opened 2026-03-24
Increase escalation path nesting depth from 3 to 5 levels

JoshuaGuy22:
The previous limit of 3 levels was insufficient for complex timezone-aware routing scenarios. For example:

Level 1: Check if priority is critical
Level 2: Check if priority is warning (inside else)
Level 3: Check if MST business hours active (inside then)
Level 4: Check if BST business hours active (inside else)

This change increases the depth from 4 to 6 (supporting 5 levels of nesting) to accommodate these more complex escalation patterns.

Fixes: Nested if_else structures failing with 'mismatch between struct and object' error


phroggyy:
Hi @JoshuaGuy22 - sorry, I'd missed seeing this PR, and have just shipped this separately in v5.35.2. Going to close this PR as a result!

PR #332, opened 2026-04-06
fix: handle 404 in catalog_entries Read for externally deleted resources

moabukar:
## What

Adds 404 handling in the `incident_catalog_entries` resource's `Read()` method so that externally deleted catalog types are removed from state instead of erroring.

## Why

If someone deletes a catalog type outside Terraform (UI, API, etc), `terraform plan` currently blows up with a generic client error. After this change it detects the 404, drops the resource from state, and plans a re-create — same as every other resource in this provider already does.

## Testing

- `go build ./...` passes
- Existing acceptance tests unaffected

### DIFF
--- a/internal/provider/incident_catalog_entries_resource.go
+++ b/internal/provider/incident_catalog_entries_resource.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


finamadrova (approved):
Yes please, this is a very nice addition to the provider.

PR #340, opened 2026-04-15
Add plan-time source_type validation and document alert source variants

thundering-herd:
## Background

The recent addition of heartbeat monitoring to the `incident_alert_source` resource (v5.35.0) wasn't reflected in the docs. The `heartbeat_options` block was implemented and working, but there was no example showing how to use it — leaving users to figure out the structure themselves.

While looking into how to add that example, a few related gaps became apparent that were worth addressing at the same time.

## What was missing and why it matters

**No examples for structurally distinct source types**

The existing example only covered the generic webhook pattern (CloudWatch). But some source types are structurally different — they require extra options blocks or have different rules around the template. Without examples, users hitting `heartbeat`, `jira`, or `http_custom` for the first time would have no guidance on what those differences are.

The decision was to document one example per structural variant rather than one per source type. With 45 source types, most of which are identical webhooks, exhaustive documentation would be noise. The vendor's own recommendation to use the Export flow for template content reinforces this — the examples cover the *shape* of the config, not every possible source.

**No plan-time validation on `source_type`**

`source_type` accepts a string but had no validation. Combined with `RequiresReplace()` (changing source type destroys and recreates the resource), a typo meant: Terraform plans a destroy+create, you approve it, apply starts, the API rejects the value with a cryptic error, and you've already destroyed the original resource.

Adding `StringOneOfValidator` catches this at `terraform plan` before any infrastructure is touched. The list of valid values is derived from `EnumValues()` — a new helper in `apischema` that reads the enum directly from the OpenAPI schema JSON already in the repo, so it stays in sync automatically when new source types are added.

**`http_custom` had no test coverage**

Of the four structurally distinct source types, three already had acceptance tests. `http_custom` was the only one missing. This adds `TestAccAlertSourceResource_HTTPCustom` to close that gap.

## Decisions made

- Listing all 45 `source_type` values inline in the schema description was considered and rejected — a wall of backtick-wrapped values in the docs table hurts readability more than it helps. The validator error message is the right place for the full list; it appears exactly when needed.
- `StringOneOfValidator` was implemented as a custom validator consistent with the existing `NonEmptyListValidator` and `RFC3339TimestampValidator` in the codebase, rather than adding `terraform-plugin-framework-validators` as a new dependency.
- `EnumValues` was added to `apischema` to return `[]string` for use by `StringOneOfValidator`. This is intentionally separate from the existing `EnumValuesDescription` in `helpers.go`, which formats enum values as a human-readable string for descriptions — two different purposes, one helper each.
- An earlier iteration added `DocstringWithEnum` to `apischema` which duplicated `EnumValuesDescription`. This was caught in review and removed — `EnumValuesDescription` in `helpers.go` remains the single implementation for formatted enum descriptions.

## Test plan

- [x] `INCIDENT_API_KEY=dummy TF_ACC=1 go test ./internal/provider -run TestAccAlertSourceResource_ValidationErrors` — passes without a real API key
- [ ] `INCIDENT_API_KEY=<key> make testacc TESTARGS="-run TestAccAlertSourceResource_Heartbeat"` — requires real API
- [ ] `INCIDENT_API_KEY=<key> make testacc TESTARGS="-run TestAccAlertSourceResource_HTTPCustom"` — requires real API
- [x] `make generate` — regenerates `docs/resources/alert_source.md` cleanly

### DIFF
--- a/docs/resources/alert_source.md
+++ b/docs/resources/alert_source.md
@@ -19,111 +19,188 @@ We'd generally recommend building alert sources in our [web dashboard](https://a
 ## Example Usage
 
 ```terraform
-## Create a basic Alert Source that receives from an SNS Topic in AWS
+## Most alert sources share the same structure: a name, a source_type, and a template
+## that describes how to map the incoming payload to an incident alert.
+##
+## We'd generally recommend building alert sources in the incident.io web dashboard
+## (https://app.incident.io/~/alerts/configuration) and using the 'Export' flow to
+## generate your Terraform, rather than hand-crafting the template. The examples below
+## cover the structural variants — source types that require extra options blocks or
+## have different template rules.
+
+## -----------------------------------------------------------------------------
+## Generic webhook source (e.g. cloudwatch, datadog, grafana, sns, …)
+## -----------------------------------------------------------------------------
+## All standard webhook-based source types follow this pattern. The template
+## title and description are JSON-encoded rich-text documents; use the Export
+## flow in the dashboard to get the exact values for your source type.
 
 resource "incident_alert_source" "cloudwatch" {
   name        = "CloudWatch Alerts"
   source_type = "cloudwatch"
+
   template = {
     title = {
       literal = jsonencode({
-        content = [
-          {
-            content = [
-              {
-                attrs = {
-                  label   = "Payload → Title"
-                  missing = false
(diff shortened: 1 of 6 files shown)


thundering-herd:
Will close this due to inactivity.

PR #353, opened 2026-05-21
Fix "inconsistent result after apply" for http_custom_options

### LINKED ISSUE #352: incident_alert_source: "inconsistent result after apply" for http_custom_options (source_type = "http")
### Summary

Creating an `incident_alert_source` with `source_type = "http"` (and `http_custom_options`) always fails on apply:

```
When applying changes to module.opnsense[...].incident_alert_source.opnsense,
provider produced an unexpected new value: .http_custom_options: was
cty.ObjectVal(map[string]cty.Value{"deduplication_key_path":cty.StringVal("$.alert_id"),
"transform_expression":cty.StringVal("...")}), but now null.
This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

### Provider version

v5.38.0 (latest)

### Root cause

The create/update API response does not echo back `http_custom_options` — it appears to be write-only (`deduplication_key_path` is even redacted as `**********` in the plan). So `AlertSourceHTTPCustomOptionsModel.FromAPI` receives `nil` and returns `nil` (`internal/provider/models/alert_source_v1_model.go`). The planned non-null `http_custom_options` then becomes `null` in the post-apply state, and the framework rejects the inconsistent result.

The provider already handles exactly this pattern for `email_options` and `auto_resolve_incident_alerts` in `Create`/`Read`/`Update` (`internal/provider/incident_alert_source_resource.go`) by restoring the planned/prior value when `FromAPI` returns nil — but the equivalent was never added for `http_custom_options` (HTTP support was added later, in #244).

### Impact

`source_type = "http"` alert sources cannot be managed by Terraform at all. `lifecycle`/`ignore_changes` does not suppress this framework-level error.

### Minimal reproduction

```hcl
resource "incident_alert_source" "test" {
  name        = "http-source"
  source_type = "http"

  template = {
    expressions = []
    title       = { literal = "..." }
    description = { literal = "..." }
    attributes  = []
  }

  http_custom_options = {
    transform_expression   = "payload"
    deduplication_key_path = "$.alert_id"
  }
}
```

`terraform apply` → inconsistent result error above.

### Fix

Mirror the existing `email_options` workaround: preserve the planned value (Create/Update) / prior-state value (Read) for `http_custom_options` when `FromAPI` returns nil. PR incoming.

allanice001:
Fixes #352.

### Problem

Creating (or updating) an `incident_alert_source` with `source_type = "http"` fails:

```
... .http_custom_options: was cty.ObjectVal(...), but now null.
```

The create/update API response does not echo back `http_custom_options` (it's write-only — `deduplication_key_path` is redacted in plans), so `AlertSourceHTTPCustomOptionsModel.FromAPI` returns nil and the planned non-null value becomes null in post-apply state.

### Fix

This mirrors the existing handling for `email_options` and `auto_resolve_incident_alerts`: when `FromAPI` returns nil for `http_custom_options`, restore the planned value in `Create`/`Update` and the prior-state value in `Read` (the latter avoids a perpetual diff). Three call sites in `internal/provider/incident_alert_source_resource.go`.

### Test

Adds `TestAccAlertSourceResource_HTTPCustomOptions`, which creates an http source with `http_custom_options` and asserts apply succeeds with no follow-up diff. The import step uses `ImportStateVerifyIgnore: ["http_custom_options"]` since the API never returns it.

### Checks run locally

- `gofmt -l` — clean
- `go build ./...` — ok
- `go vet ./internal/provider/` — ok
- Acceptance test requires a live incident.io API key (`TF_ACC`); not run locally.

---

> [!NOTE]
> **Medium Risk**
> Touches `Create`/`Read`/`Update` state reconciliation for `incident_alert_source`, which can affect diffs and state for HTTP sources if the preservation logic is incorrect. Scope is limited to a single write-only attribute and is covered by a new acceptance regression test.
> 
> **Overview**
> Fixes a Terraform **"inconsistent result after apply"** / perpetual diff for `incident_alert_source` when `http_custom_options` is set on `source_type = "http"`.
> 
> The provider now preserves `http_custom_options` from the plan (on `Create`/`Update`) or prior state (on `Read`) when the API response omits this *write-only* field, mirroring existing handling for `email_options` and `auto_resolve_incident_alerts`. Adds an acceptance regression test for HTTP sources and notes the fix in `CHANGELOG.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit abe007b78223706c0df9324f8972acc8b90c9612. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Fix "Provider produced inconsistent result after apply" when creating or updating an `incident_alert_source` with `http_custom_options` (the API does not echo back the write-only `http_custom_options`, so the planned/prior value is now preserved)
+
 ## v5.38.0
 
 - Add `incident_schedule_sync_target` and `incident_schedule_sync_rule` resources for syncing schedules to Slack user groups
--- a/internal/provider/incident_alert_source_resource.go
+++ b/internal/provider/incident_alert_source_resource.go
@@ -513,6 +513,7 @@ func (r *IncidentAlertSourceResource) Create(ctx context.Context, req resource.C
 	// Save the planned values before overwriting with API response.
 	planAutoResolveIncidentAlerts := data.AutoResolveIncidentAlerts
 	planEmailOptions := data.EmailOptions
+	planHTTPCustomOptions := data.HTTPCustomOptions
 
 	data = models.AlertSourceResourceModel{}.FromAPI(result.JSON200.AlertSource)
 
@@ -529,6 +530,13 @@ func (r *IncidentAlertSourceResource) Create(ctx context.Context, req resource.C
 	// doesn't see an inconsistent result.
 	if planEmailOptions != nil && data.EmailOptions == nil {
 		data.EmailOptions = planEmailOptions
+	}
+
+	// http_custom_options is write-only: the API never echoes it back, so
+	// FromAPI returns nil. Restore the planned value so Terraform doesn't see
+	// an inconsistent result for http sources.
+	if planHTTPCustomOptions != nil && data.HTTPCustomOptions == nil {
+		data.HTTPCustomOptions = planHTTPCustomOptions
 	}
 
 	if data.SourceType.ValueString() == "heartbeat" && data.Template != nil {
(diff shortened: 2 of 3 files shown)


mbfisher (approved):
Good to go!

allanice001:
merge conflict resolved in changelog

mbfisher (approved):

isaacseymour:
Hey! This is because `http_custom_options` is only supported when `source = "http_custom"`.

I'm going to change the API to reject requests where the source and options are incompatible, but you'll need to update your terraform config here.

PR #356, opened 2026-06-01
ONC-11917: Fix escalation path crash on unknown path/working_hours values

isaacseymour:
The `incident_escalation_path` resource crashes at plan time with a `Value Conversion Error: Received unknown value, however the target type cannot handle unknown values` whenever any part of `path` or `working_hours` is unknown — for example `path = local.path_templates[var.path_template]`, or a target/working-hours field that references a not-yet-created resource. This regressed in v5.37.0 (the escalation-path schedule modes change) and affects 5.37.0, 5.38.0 and 5.38.1, blocking any plan that builds escalation paths dynamically.

The root cause was in `ValidateConfig`, which called `req.Config.Get` to decode the whole config into `IncidentEscalationPathResourceModel`. That model represents `Path` and `WorkingHours` as plain Go slices (`[]IncidentEscalationPathNode`, `[]models.IncidentWeekdayIntervalConfig`), and plain slices cannot hold an unknown list or element, so the decode errors out before any validation runs. `ValidateConfig` is the only place that decodes config-with-unknowns: Create/Update use `Plan.Get` at apply time when values are known, and Read uses state.

The fix stops decoding the full model in `ValidateConfig`. We now pull just the `path` attribute as a `types.List` via `GetAttribute` and walk it recursively on the framework `attr.Value` types, descending only into elements and nested objects that are known and skipping anything unknown or null. This matters because even when the `path` list itself is known, an individual element or a leaf such as `schedule_mode`/`selected_rota_id` can still be unknown — decoding those into a plain struct would crash. The existing `selected_rota_id`/`schedule_mode` consistency checks are preserved (and still bail out when those leaves are unknown), so validation behaviour is unchanged for fully-known configs. Added acceptance tests cover both failure modes: a `path` sourced from a variable-indexed local, and a `working_hours` entry referencing a known-after-apply attribute.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Fix `incident_escalation_path` crashing at plan time with a "cannot handle unknown values" error when `path` or `working_hours` contain values that are unknown until apply (e.g. `path = local.path_templates[var.path_template]`)
+
 ## v5.38.1
 
 - Mark `incident_schedule_sync_target` and `incident_schedule_sync_rule` resources as managed by Terraform
--- a/internal/provider/incident_escalation_path_resource.go
+++ b/internal/provider/incident_escalation_path_resource.go
@@ -389,13 +389,26 @@ func (r *IncidentEscalationPathResource) Configure(ctx context.Context, req reso
 }
 
 func (r *IncidentEscalationPathResource) ValidateConfig(ctx context.Context, req resource.ValidateConfigRequest, resp *resource.ValidateConfigResponse) {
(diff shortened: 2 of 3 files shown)


isaacseymour:
Closing in favour of the full fix for ONC-11917. A feasibility spike showed the larger refactor — converting `path`/`then_path`/`else_path` and `working_hours`/`weekday_intervals` to `types.List` so the model can carry unknowns end-to-end — is more tractable than expected (the recursive object-attr-types concern didn't materialise). We're landing that instead, since it eliminates the whole bug class rather than just hardening `ValidateConfig`. New PR to follow.

PR #357, opened 2026-06-01
ONC-11943: Fix inconsistent-result error for engine literals containing HTML chars

isaacseymour:
Engine param-binding `literal` values that contain HTML characters (`>`, `<`, `&`) and were supplied by an encoder that does not HTML-escape — CDKTF `JSON.stringify`, `file()`, heredocs, hand-written JSON — failed with `Provider produced inconsistent result after apply`, and in some cases produced a perpetual diff. HCL's `jsonencode` HTML-escapes by default, which is why the majority of users never hit this and why it surfaced for a CDKTF user.

The cause is an asymmetry between the two code paths in `internal/provider/models/engine.go`. The inbound plan → API path sends the literal verbatim, so the planned value Terraform stores is the raw config string. The outbound API → state path re-encodes the literal, which can change its bytes (key ordering, and previously HTML escaping). Terraform then compares the planned literal against the post-apply state literal byte-for-byte, so a `>` planned value versus a `>` applied value is treated as a mismatch. This is a no-win toggle: turning `SetEscapeHTML` off helps raw-string users but breaks `jsonencode` users in the opposite direction, and vice versa — exactly the round-trip in ONC-7057 / ONC-7504 — because only one side is ever normalised.

The fix is a reusable semantic-equality string type, `jsontypes.NormalizedString` (new `internal/provider/jsontypes` package). Its `StringSemanticEquals` parses both the old and new values: if both are valid JSON it compares their canonical forms (key-sorted and escaping-insensitive, so `>` and `>` compare equal); if either is not JSON it falls back to exact string equality, which is essential because a literal can also be a plain reference or string. This is applied to the `literal` field everywhere it is defined — the resource schema in `ParamBindingValueAttributes()`, the data-source schemas (`engine_datasource.go` and the inline escalation-path data source), the `IncidentEngineParamBindingValue.Literal` model field, and every `types.StringType` object-attribute-type entry for `literal` in `alert_route_v1_model.go` so the schema and value types stay consistent. The form written into state still keeps HTML escaping on, so it remains byte-stable for the dominant `jsonencode` population and for `ImportStateVerify`; the escaping-insensitive comparison is used only for semantic equality. Unit tests cover the equality logic (identical / reordered / whitespace / HTML-escaped / non-JSON / null-unknown cases) and an acceptance test reproduces the original scenario with a raw heredoc literal alongside a parallel `jsonencode` case.


paprikati (approved):
I think this is real smart, and if it isn't for some reason I can't figure out we can revert easily.

PR #361, opened 2026-06-15
Model engine `expressions` as ListNestedAttribute (fixes #360)

### LINKED ISSUE #360: Alert source/route/workflow `expressions` (SetNestedAttribute) fails `terraform validate` when some expressions omit optional `else_branch` (incl. Export-to-Terraform output)
### Summary

`incident_alert_source.template.expressions` (and the shared `expressions` attribute on `incident_alert_route` and `incident_workflow`) is modeled as a `SetNestedAttribute` whose element object has an **optional** `else_branch`. When a configuration contains a mix of expressions — some with `else_branch` and some without — `terraform validate` fails before any API call with:

```
Error: Incorrect attribute value type
  Inappropriate value for attribute "template": element types must all match
  for conversion to set.
```

This is exactly the shape that the incident.io dashboard's **"Export to Terraform"** feature produces (parse expressions have no `else_branch`; a `branches` expression has one), so the exported HCL does not pass `terraform validate` as-is. It's also the natural representation of any imported alert source that has both kinds of expression, which forces a workaround: adding a placeholder `else_branch` to the expressions that lack one (so all set elements share a single type) — but the API doesn't persist those placeholders, producing permanent plan drift that then has to be silenced with `ignore_changes` on the expressions.

### Provider / Terraform versions

- Provider `incident-io/incident` v5.39.0 (also present on `master`)
- Terraform v1.2.8 (this is core set-conversion behavior, not version-specific)

### Minimal reproduction

```hcl
terraform {
  required_providers {
    incident = { source = "incident-io/incident", version = "5.39.0" }
  }
}

provider "incident" {
  api_key = "not-used-by-validate"
}

resource "incident_alert_source" "repro" {
  name        = "repro"
  source_type = "http"
  template = {
    title       = { literal = jsonencode({ type = "doc", content = [] }) }
    description = { literal = jsonencode({ type = "doc", content = [] }) }
    attributes  = []

    expressions = [
      # parse expression — NO else_branch
      {
        label          = "A"
        reference      = "a"
        root_reference = "payload"
        operations = [{
          operation_type = "parse"
          parse = { returns = { array = false, type = "String" }, source = "$.a" }
        }]
      },
      # expression WITH else_branch
      {
        label          = "B"
        reference      = "b"
        root_reference = "payload"
        else_branch    = { result = { value = { literal = "x" } } }
        operations = [{
          operation_type = "parse"
          parse = { returns = { array = false, type = "String" }, source = "$.b" }
        }]
      },
    ]
  }
}
```

`terraform init && terraform validate` → the error above. Removing `else_branch` from expression "B" (so all elements share a type) makes it validate; that's not an option when the real config legitimately needs `else_branch` on some expressions only.

### Production-like example (sanitized)

The minimal case above is deliberately tiny. For context, this is the shape of a real CloudWatch alert source — title/description templates, a mix of literal and reference attribute bindings, several `parse` expressions with no `else_branch`, and a `branches` expression that derives a priority and carries an `else_branch` default. (IDs, attribute names, and match values are sanitized placeholders.) It fails `terraform validate` with the same error; the same provider block as above applies.

```hcl
resource "incident_alert_source" "service_alarms" {
  name        = "Service Alarms"
  source_type = "cloudwatch"

  template = {
    title = {
      literal = jsonencode({
        type = "doc"
        content = [{
          type = "paragraph"
          content = [{
            type  = "varSpec"
            attrs = { label = "Payload → Title", missing = false, name = "title" }
          }]
        }]
      })
    }

    description = {
      literal = jsonencode({
        type = "doc"
        content = [{
          type    = "paragraph"
          content = [{ type = "text", text = "See the alarm payload above." }]
        }]
      })
    }

    # A mix of reference and literal bindings, as a real source typically has.
    attributes = [
      {
        alert_attribute_id = "01AAAAAAAAAAAAAAAAAAAAPRIO"
        binding = {
          value          = { reference = "expressions[\"priority\"]" }
          merge_strategy = "last_wins"
        }
      },
      {
        alert_attribute_id = "01AAAAAAAAAAAAAAAAAAAATEAM"
        binding = {
          value          = { literal = "01AAAAAAAAAAAAAAAAATEAMENT" }
          merge_strategy = "first_wins"
        }
      },
      {
        alert_attribute_id = "01AAAAAAAAAAAAAAAAAAAAAENV"
        binding = {
          value          = { reference = "expressions[\"environment\"]" }
          merge_strategy = "first_wins"
        }
      },
    ]

    expressions = [
      # Parse expression — NO else_branch.
      {
        label          = "Environment"
        reference      = "environment"
        root_reference = "payload"
        operations = [{
          operation_type = "parse"
          parse = {
            source  = <<-EOT
              function extractAttribute(params) {
                return $.AlarmName.split('-').slice(0, 2).join('-');
              }
              extractAttribute();
            EOT
            returns = { type = "String", array = false }
          }
        }]
      },
      # Parse expression — NO else_branch.
      {
        label          = "Account ID"
        reference      = "account_id"
        root_reference = "payload"
        operations = [{
          operation_type = "parse"
          parse = {
            source  = <<-EOT
              function extractAttribute(params) {
                return $.AWSAccountId;
              }
              extractAttribute();
            EOT
            returns = { type = "Number", array = false }
          }
        }]
      },
      # Branches expression — HAS else_branch (default priority).
      {
        label          = "Priority"
        reference      = "priority"
        root_reference = "."
        else_branch = {
          result = { value = { literal = "01AAAAAAAAAAAAAAADEFAULTPR" } }
        }
        operations = [{
          operation_type = "branches"
          branches = {
            returns = { type = "CatalogEntry[\"AlertPriority\"]", array = false }
            branches = [
              {
                condition_groups = [{
                  conditions = [{
                    subject   = "title"
                    operation = "contains_one_of"
                    param_bindings = [{
                      array_value = [{ literal = "prod" }]
                    }]
                  }]
                }]
                result = { value = { literal = "01AAAAAAAAAAAAAAAAHIGHPRIO" } }
              },
            ]
          }
        }]
      },
    ]
  }
}
```

joncon-hivewatch:
## Overview

`expressions` (on `incident_alert_source`, `incident_alert_route` and `incident_workflow`) is a `SetNestedAttribute` whose element has an optional `else_branch`. Any config that mixes expressions with and without `else_branch` fails `terraform validate` with `element types must all match for conversion to set` — including the HCL your dashboard's *Export to Terraform* emits. This changes `expressions` to a `ListNestedAttribute` so those configs validate.

Fixes #360.

## Detailed Technical Description of Change

- A Terraform `set` requires a single homogeneous element type. When the config has object literals where the optional `else_branch` is present on some elements and absent on others, the tuple→set conversion can't unify them and fails at decode time, before any API call.
- `ExpressionsAttribute()` in `internal/provider/models/engine.go` is changed from `schema.SetNestedAttribute` to `schema.ListNestedAttribute`. This is shared by all three resources. The sibling `operations` attribute is already a `ListNestedAttribute` and doesn't have the problem; expressions are referenced by their `reference` key rather than by position; and the alert-source resource's existing plan modifier already traverses expressions via `.AtListIndex(...)`.
- Scope is intentionally minimal: only the resource attribute changes. The read-only data-source equivalent (`ExpressionsDataSourceAttribute`) is computed and unaffected, so it's left as-is.
- Tradeoff worth a maintainer eye: a list is order-sensitive, so `Read` should return expressions in a stable order to avoid spurious diffs. In practice import and *Export to Terraform* already emit a consistent order.

## Testing Approach and Results

- **Reproduction (before/after):** the minimal config in #360, and a larger production-shaped config (multiple `parse` expressions without `else_branch` plus a `branches` expression with one), both fail `terraform validate` on the released provider and pass on this build.
- **Regression test:** added `TestAccAlertSourceResource_Issue360_MixedElseBranch`, an acceptance test that plans an alert source whose expressions mix present/absent `else_branch` (runs in CI against the test account).
- **Static checks:** `go vet ./...`, `go build ./...`, and the `internal/provider/models` unit tests all pass.
- **Plan behavior:** validated against a realistic existing configuration — switching the schema to a list produced no spurious reorder/`else_branch` diff on plan.

## Documentation

- Regenerated resource docs via `tfplugindocs` (`expressions`: Attributes Set → List): `docs/resources/alert_source.md`, `docs/resources/alert_route.md`, `docs/resources/workflow.md`.

### DIFF
--- a/docs/resources/alert_route.md
+++ b/docs/resources/alert_route.md
@@ -246,7 +246,7 @@ resource "incident_alert_route" "service_alerts" {
 - `condition_groups` (Attributes List) Groups of prerequisite conditions. All conditions in at least one group must be satisfied (see [below for nested schema](#nestedatt--condition_groups))
 - `enabled` (Boolean) Whether this alert route is enabled or not
 - `escalation_config` (Attributes) (see [below for nested schema](#nestedatt--escalation_config))
-- `expressions` (Attributes Set) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
+- `expressions` (Attributes List) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
 - `incident_config` (Attributes) (see [below for nested schema](#nestedatt--incident_config))
 - `incident_template` (Attributes) (see [below for nested schema](#nestedatt--incident_template))
 - `is_private` (Boolean) Whether this alert route is private. Private alert routes will only create private incidents from alerts.
--- a/docs/resources/alert_source.md
+++ b/docs/resources/alert_source.md
@@ -160,7 +160,7 @@ Required:
 
 - `attributes` (Attributes Set) Attributes to set on alerts coming from this source, with a binding describing how to set them. (see [below for nested schema](#nestedatt--template--attributes))
 - `description` (Attributes) (see [below for nested schema](#nestedatt--template--description))
-- `expressions` (Attributes Set) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--template--expressions))
+- `expressions` (Attributes List) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--template--expressions))
 - `title` (Attributes) (see [below for nested schema](#nestedatt--template--title))
 
 Optional:
--- a/docs/resources/workflow.md
+++ b/docs/resources/workflow.md
@@ -83,7 +83,7 @@ resource "incident_workflow" "autoassign_incident_lead" {
 
 - `condition_groups` (Attributes List) Groups of prerequisite conditions. All conditions in at least one group must be satisfied (see [below for nested schema](#nestedatt--condition_groups))
 - `continue_on_step_error` (Boolean) Whether to continue executing the workflow if a step fails
-- `expressions` (Attributes Set) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
+- `expressions` (Attributes List) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
 - `include_private_incidents` (Boolean) Whether to include private incidents
 - `name` (String) Name provided by the user when creating the workflow
 - `once_for` (List of String) This workflow will run 'once for' a list of references
--- a/internal/provider/incident_alert_source_resource_test.go
+++ b/internal/provider/incident_alert_source_resource_test.go
@@ -1392,3 +1392,77 @@ resource "incident_alert_source" "test" {
 		TeamTypeName: teamTypeName(),
 	})
 }
+
+// TestAccAlertSourceResource_Issue360_MixedElseBranch is a regression test for
+// #360: expressions that mix present and absent else_branch must plan without
+// "element types must all match for conversion to set".
+func TestAccAlertSourceResource_Issue360_MixedElseBranch(t *testing.T) {
+	resource.Test(t, resource.TestCase{
+		PreCheck:                 func() { testAccPreCheck(t) },
+		ProtoV6ProviderFactories: testAccProtoV6ProviderFactories,
+		Steps: []resource.TestStep{
+			{
+				Config:             testAccAlertSourceResourceConfigMixedElseBranch(),
+				PlanOnly:           true,
+				ExpectNonEmptyPlan: true,
+			},
+		},
+	})
+}
+
+func testAccAlertSourceResourceConfigMixedElseBranch() string {
+	return fmt.Sprintf(`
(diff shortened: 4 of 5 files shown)


joncon-hivewatch:
One note on the ordering tradeoff mentioned above: this PR keeps things minimal with a plain `ListNestedAttribute`. If you'd prefer to preserve set-like order-insensitivity (so reordering expressions, or the API returning them in a different order, doesn't produce a diff), an alternative would be to keep the list but add order-insensitive diffing via a semantic-equality custom type / sort-by-`reference` plan modifier — mirroring the `jsontypes.NormalizedJSONOrString` pattern already used for JSON literal fields. Happy to take that approach instead if you'd rather go that route.

benfdking:
I am closing this because of https://github.com/incident-io/terraform-provider-incident/issues/360#issuecomment-5422189720. Please do reopen the issue if my analysis is incorrect!

PR #368, opened 2026-06-25
Fix expressions `terraform validate` failure when some expressions omit `else_branch`

isaacseymour:
## Summary

Fixes RESP-17992. The shared `expressions` attribute on `incident_alert_source` (`template.expressions`), `incident_alert_route` and `incident_workflow` was modelled as a `SetNestedAttribute` whose element object has an **optional** `else_branch`.

On the Terraform 1.2.x versions this provider targets (see the CI matrix), a config containing a mix of expressions — some with `else_branch`, some without — fails at **validate time, before any API call**, with:

```
Inappropriate value for attribute "template": element types must all match
for conversion to set.
```

This is exactly the shape the dashboard's **Export to Terraform** feature emits (parse expressions have no `else_branch`; a `branches` expression carries one), so the exported HCL doesn't pass `terraform validate` as-is. The previous workaround was to add a placeholder `else_branch` to expressions that lack one and then silence the resulting drift with `ignore_changes`.

## Root cause

Terraform core decodes the config tuple and converts it to the attribute's type. For a **set**, this goes through a tuple→set conversion that does not honour optional attributes on Terraform 1.2.x, so two elements with structurally different object shapes (one with `else_branch`, one without) are rejected. Confirmed against real Terraform binaries: it reproduces on **1.2.8** and is fixed in core by **1.9.8**. Since the provider still targets 1.2.x, the fix must live in the provider.

## Fix

Model `expressions` as a `ListNestedAttribute` instead of a `SetNestedAttribute`. Lists are decoded element-by-element and tolerate the heterogeneous objects, so the mixed config validates on all Terraform versions.

### Keeping ordering stable (the reason for the v5.21.1 revert)

A previous list→set change was reverted in **v5.21.1** because the API does not return `expressions` in a stable order; a plain list resurfaced as `Provider produced inconsistent result after apply` and perpetual diffs (this is exactly what Bugbot flagged, and it showed up as a CI failure in `TestAccAlertSourceResource_Issue342_SpuriousMergeStrategy`).

To keep the list but avoid that drift, this PR adds `IncidentEngineExpressions.ReorderToMatch`, which realigns the API response to the planned/prior order keyed by the unique `reference` field (any extras are appended in API order). It's wired into:

- `incident_alert_source` — Create/Read/Update (reorder against plan on create/update, against prior state on read).
- `incident_alert_route` — inside the existing `FromAPIWithPlan` (reorder against the threaded plan/state).
- `incident_workflow` — Create/Read/Update (reorder against plan/state around `buildModel`).

The data source (`incident_alert_sources`) is left as a set: it's read-only and populated from the API, so neither the config-conversion nor the ordering path applies.

## Testing

- `TestAccAlertSourceResource_MixedElseBranchExpressions` — a `PlanOnly` regression test mirroring the Export-to-Terraform output. Verified against Terraform 1.2.8: fails with the `element types must all match for conversion to set` error on the old `Set` schema; passes with the `List` fix. The failure is at config-validation time, so it needs no API credentials.
- `TestIncidentEngineExpressions_ReorderToMatch` — unit test covering reordering, appending unmatched expressions, and the empty cases.

Also updated the generated docs (`expressions` now reads `(Attributes List)`) and added a CHANGELOG entry noting the possible one-time reorder on upgrade.

## Notes / risk

Changing an attribute from set to list is a schema type change. Stored state and config both serialise the value as a JSON array, so re-reading existing state as a list works structurally. Existing resources may see a one-time reordering of `expressions` in the first plan after upgrade (expected, resolves on apply); thereafter the `ReorderToMatch` realignment keeps the list stable against API reordering.

Linear Issue: [RESP-17992](https://linear.app/incident-io/issue/RESP-17992/alert-sourcerouteworkflow-expressions-setnestedattribute-fails)

<div><a href="https://cursor.com/agents/bc-376a1e09-1d87-42a2-9837-9b7ea90d13d3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-376a1e09-1d87-42a2-9837-9b7ea90d13d3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Fix `terraform validate` failing with `element types must all match for conversion to set` when the `expressions` attribute (on `incident_alert_source`, `incident_alert_route` and `incident_workflow`) contains a mix of expressions with and without an `else_branch`. This is the shape the dashboard's "Export to Terraform" feature produces. `expressions` is now a list rather than a set, which tolerates the heterogeneous elements without needing placeholder `else_branch` workarounds or `ignore_changes`.
+
 ## v5.40.0
 
 - Add an optional `rotation_id` to `incident_schedule_sync_rule`, scoping a sync rule to a single rotation on the schedule. Omit it to sync all rotations (the previous behaviour); to feed a Slack user group from several rotations, create one rule per rotation pointing at the same target.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Unreleased
 
-- Fix `terraform validate` failing with `element types must all match for conversion to set` when the `expressions` attribute (on `incident_alert_source`, `incident_alert_route` and `incident_workflow`) contains a mix of expressions with and without an `else_branch`. This is the shape the dashboard's "Export to Terraform" feature produces. `expressions` is now a list rather than a set, which tolerates the heterogeneous elements without needing placeholder `else_branch` workarounds or `ignore_changes`.
+- Fix `terraform validate` failing with `element types must all match for conversion to set` when the `expressions` attribute (on `incident_alert_source`, `incident_alert_route` and `incident_workflow`) contains a mix of expressions with and without an `else_branch`. This is the shape the dashboard's "Export to Terraform" feature produces. `expressions` is now a list rather than a set, which tolerates the heterogeneous elements without needing placeholder `else_branch` workarounds or `ignore_changes`. To avoid the ordering drift that previously forced this back to a set (v5.21.1), the provider now realigns the API's expression response to the planned/prior order by reference, so there's no spurious diff or "inconsistent result after apply". Existing resources may see a one-time reordering of `expressions` in the first plan after upgrading, which is expected and resolves on apply.
 
 ## v5.40.0
 
--- a/docs/resources/alert_route.md
+++ b/docs/resources/alert_route.md
@@ -246,7 +246,7 @@ resource "incident_alert_route" "service_alerts" {
 - `condition_groups` (Attributes List) Groups of prerequisite conditions. All conditions in at least one group must be satisfied (see [below for nested schema](#nestedatt--condition_groups))
 - `enabled` (Boolean) Whether this alert route is enabled or not
 - `escalation_config` (Attributes) (see [below for nested schema](#nestedatt--escalation_config))
-- `expressions` (Attributes Set) The expressions to be prepared for use by steps and conditions (see [below for nested schema](#nestedatt--expressions))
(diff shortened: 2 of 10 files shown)


benfdking:
Closing this unmerged — the bug it fixes no longer reproduces on any Terraform we support.

The PR's premise was that the tuple→set conversion bug is present in core 1.2.8 and fixed by 1.9.8, so "since the provider still targets 1.2.x, the fix must live in the provider". That stopped being true in 08397ae ("Require Terraform 1.14, test against OpenTofu"): the CI matrix is now Terraform 1.14.\*/1.15.\* plus OpenTofu 1.12.5, and README.md/docs/index.md state that from v6.0.0 we support Terraform 1.14+ and won't be fixing issues that only reproduce on older CLIs.

Verified rather than inferred: built master (expressions is still a `SetNestedAttribute`, internal/provider/models/engine.go:449), pointed `dev_overrides` at it and ran the exact minimal repro from RESP-17992:

- Terraform 1.15.8 → `Success! The configuration is valid`
- OpenTofu 1.12.5 → `Success! The configuration is valid`

So the validate failure is gone on both supported CLIs, and the placeholder-`else_branch` + `ignore_changes` workaround goes with it (it only ever existed to dodge the validate error).

The costs of merging are unchanged, though: a set→list schema change across three resources, a one-time reorder in every existing user's first plan, `ReorderToMatch` to maintain indefinitely, and re-entering the failure mode that forced the v5.21.1 revert — which is what Bugbot flagged here. `TestAccAlertSourceResource_MixedElseBranchExpressions` would now pass against the set schema too, so it wouldn't be guarding anything.

Two other things worth noting for anyone reading this later:

- The branch is conflicting: `internal/provider/models/alert_route_v1_model.go`, one of the files it patches, was deleted by 1995390 ("Support the v3 alert routes API").
- The v3 beta path already models expressions as list-nested blocks (`models.NamedExpressionBlock`, used by `alert_source_beta_resource.go`), so heterogeneous elements can't arise there by construction — the direction this PR wanted, arrived at another way.

`ReorderToMatch` is worth remembering if expressions ever do move to a list for some other reason. The remaining set-correlation bugs aren't ordering: ONC-9362 was cancelled, and ONC-12476 / PD-20912 are API-serializer echoes.

PR #370, opened 2026-06-26
Support the v3 alert routes API in the incident_alert_route resource

acroz:
## What

Rather than ship a **separate** `incident_alert_route_v3` resource, this makes the **existing** `incident_alert_route` resource support both the v2 and v3 alert routes APIs, switched by whether the new top-level `grouping_config` block is set.

- **v2 (default):** unchanged — `channel_config`, `message_template`, `incident_template`, and grouping fields nested under `incident_config`, served by the v2 API.
- **v3:** set `grouping_config` to opt in. Grouping moves into `grouping_config`, `channel_config` + `message_template` collapse into `message_config`, the incident template nests under `incident_config.template`, and `escalation_config` / `message_config` gain an explicit `enabled` flag. Served by the v3 API.

The v2-only fields are **deprecated** (each emits a warning pointing at its v3 home) but keep working. Because CRUD dispatches on the schema in use — `grouping_config` in the plan, or as recorded in state for read/delete — an existing route can be **migrated in place**: add the v3 blocks and drop the deprecated fields in one change, and Terraform issues an update, not a replacement. No resource-type change is required.

This replaces the earlier approach of adding a distinct `incident_alert_route_v3` resource (which would have forced a `state rm` + `import` to migrate). Most of that work is reused — the v3 model/payload/engine-bridging logic now hangs off the merged `AlertRouteResourceModel`.

## Schema sync

Bumps the embedded public schema to core master + [#63200](https://github.com/incident-io/core/pull/63200), which:
- adds the explicit `enabled` flag to v3 `escalation_config` and `message_config` (with `auto_cancel_escalations` now optional), gating their nested fields; and
- renames v3 `message_config.message_template` → `template`.

`ValidateConfig` mirrors the API's gating both ways (enabled ⇒ required fields present; disabled ⇒ gated fields unset), including `when_alert_joins_group` only being valid when both escalation and grouping are enabled — surfaced at plan time instead of as a 422.

## Review comments

Oli's comments all landed on the **generated** `docs/resources/alert_route_v3.md`, so the fixes went to the real sources:
- "behaviour when configured via `incident_alert_route` / only usable once migrated" and the repetition note → rewritten resource `MarkdownDescription` (documents the dual-schema behaviour and in-place migration, de-duplicated).
- "when the alert is resolved" wording (×2) → fixed in the example `resource.tf` (the source the docs render from).

The standalone v3 example and generated doc are removed; the merged example now shows the v3 schema with the `enabled` flags and `template`.

## Testing

- `go build`, `go vet`, `gofmt`, and the alert-route model unit tests pass (round-trip incl. the new `enabled`-gated omission, null-vs-empty, import-leaves-null).
- Acceptance tests retargeted from `incident_alert_route_v3` to `incident_alert_route` (v3-mode scenarios).
- Acceptance tests require a live org (`TF_ACC`) and were not run here. In-place v2→v3 migration is implemented via mode dispatch on the same route ID; worth a manual `terraform apply` (v2 → v3) to confirm it updates rather than replaces before release.

## Manual verification

* `terraform import` of a route with an org that's migrated stores a v3 schema alert route ✅ 
* * `terraform import` of a route with an org that's not migrated stores a v2 schema alert route ✅ 
* Trying to use the v3 alert route schema with an org which is not migrated returns a suitable error ✅ 

---

> [!NOTE]
> **Medium Risk**
> Changes how alert routes are configured and which API version is used, affecting grouping, notifications, escalations, and incident creation; migration is in-place but schema mistakes could alter production alert handling.
> 
> **Overview**
> **`incident_alert_route` now supports both v2 and v3 alert route APIs in one resource.** Setting top-level **`grouping_config`** opts into v3 and routes CRUD through the v3 API; omitting it keeps the existing v2 behavior.
> 
> The v3 shape moves grouping to **`grouping_config`**, merges **`channel_config`** and **`message_template`** into **`message_config`** (with **`template`**), nests the incident template under **`incident_config.template`**, and adds **`escalation_config.when_alert_joins_group`**. Legacy v2 fields remain usable but are **deprecated** with warnings that point to their v3 replacements. Existing routes can be **migrated in place** (add v3 blocks, remove deprecated fields) as an update rather than a replacement.
> 
> **Plan-time behavior** adds mode-aware plan modifiers so v2-only and v3-only computed attributes (e.g. **`auto_relate_grouped_alerts`**, **`when_alert_joins_group`**) plan as null when they do not apply, avoiding perpetual “known after apply” diffs and v2↔v3 migration churn. **`ValidateConfig`** mirrors API gating for v3 **`enabled`** flags and related fields.
> 
> **Import** probes the v3 API first and falls back to v2 when the org is not migrated (`api_not_yet_available`). Docs, examples, and the embedded OpenAPI client/schema are updated for the v3 field renames and related API changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f0c86b1e226b44c99bb529259c36fd041bd2920d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Follow-ups (separate PRs)

- **Docs:** add a "Migrating from the v2 to the v3 schema" guide for `incident_alert_route` (which fields move where, that migration is an in-place update, and that migrated organisations import as v3). Deliberately not included here.


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- Add the `incident_alert_route_v3` resource, supporting the v3 alert routes API. The existing `incident_alert_route` resource (v2 API) is unchanged, so this is an additive, non-breaking change. v3 moves grouping into a dedicated, required `grouping_config` (with `enabled`, `group_keys`, `window_seconds`, and a new `window_type` of `rolling`/`fixed`); collapses `channel_config` and the top-level `message_template` into a required `message_config` with `destinations`; nests the incident template under `incident_config.template`; adds an optional `when_alert_joins_group` to `escalation_config`; and drops the incident template's `workspace` binding. To migrate, switch the resource type and restructure your configuration accordingly.
```
```suggestion
- Add the `incident_alert_route_v3` resource, supporting the v3 alert routes API. The existing `incident_alert_route` resource is unchanged, and continues to work via the v2 alert route API. v3 moves grouping into a dedicated `grouping_config` section, combines `channel_config` and `message_template` into `message_config`, and nests the incident template under `incident_config.template`. To migrate, switch the resource type and restructure your configuration accordingly.
```

docs/resources/alert_route_v3.md
```diff
@@ -0,0 +1,1209 @@
...
+
+# incident_alert_route_v3 (Resource)
+
+Configure your alert routes in incident.io.
+
+Alert routes define how alerts from different sources are processed, grouped, and routed to the right teams and people.
+
+This resource uses the v3 alert routes API, which configures grouping via a dedicated `grouping_config`, combines channels and message templates into `message_config`, and nests the incident template under `incident_config`. For the previous API, see `incident_alert_route`.
+
+We'd generally recommend building alert routes in our [web dashboard](https://app.incident.io/~/alerts/configuration), and using the 'Export' flow to generate your Terraform, as it's easier to see what you've configured. You can also make changes to an existing alert route and copy the resulting Terraform without persisting it.
```
sbmooc: Repetition?

docs/resources/alert_route_v3.md
```diff
@@ -0,0 +1,1209 @@
...
+      value = {
+        literal = "01KHJTJR4FZJJZQ6G02EBFJAAY"
+      }
+    }
+  }
+
+  // Used to configure which escalation paths and/or users who should be notified when an alert is received
+  // and the conditions under which they should be notified
+  // auto_cancel_escalations is used to specify whether or not the escalation should be automatically cancelled
+  // upon receiving a 'resolved' notification for the alert that triggered the escalation
```
sbmooc: "when the alert is resolved?" instead?

docs/resources/alert_route_v3.md
```diff
@@ -0,0 +1,1209 @@
...
+    // subsequent alert joins an existing group.
+    when_alert_joins_group = {
+      mode                 = "on_priority_increase"
+      grace_period_seconds = 60
+    }
+  }
+
+  // Used to configure the incident creation settings for the alert route
+  // auto_decline_enabled is used to specify whether or not triage incidents should be automatically declined
+  // when a resolved notification is received for the alert that triggered the incident
```
sbmooc: ditto

docs/resources/alert_route_v3.md
```diff
@@ -0,0 +1,1209 @@
+---
+# generated by https://github.com/hashicorp/terraform-plugin-docs
+page_title: "incident_alert_route_v3 Resource - terraform-provider-incident"
+subcategory: ""
+description: |-
+  Configure your alert routes in incident.io.
```
sbmooc: I wonder if we want to have a few lines in here talking about how it will behave when you create an alert route using `incident_alert_route` config? And also that they can only use this route when they're fully migrated?

sbmooc (approved):
Definitely worth running past someone who has more TF experience, but looks ok to me.

PR #375, opened 2026-07-01
Support the v3 alert routes API in the incident_alert_route resource

acroz:
## What
`incident_alert_route` now supports both the v2 and v3 alert routes APIs in one resource, selected by whether the top-level `grouping_config` block is set. v3 splits grouping out of `incident_config` into `grouping_config`, merges `channel_config` + `message_template` into `message_config`, and nests the incident template under `incident_config.template`. The v2-only fields are deprecated, each with a warning pointing at its v3 location.

- CRUD dispatches on the schema in use (config for create/update, state for read/delete).
- Import probes the v3 API and falls back to v2 for organisations not yet migrated to the new alert grouping engine (`api_not_yet_available`).
- Mode-gated `ValidateConfig` and mode-aware plan modifiers keep plans clean and surface v2/v3 mistakes at plan time.

## Follow-ups (separate PRs)

- **Docs:** add a "Migrating from the v2 to the v3 schema" guide for `incident_alert_route`. Deliberately not included here.

---

> [!NOTE]
> **Medium Risk**
> Changes how alert grouping, escalations, and notifications are configured; mis-migration could alter on-call behavior, though validation and deprecation warnings reduce that risk.
> 
> **Overview**
> **`incident_alert_route`** now supports a revised configuration shape alongside the existing one. Presence of top-level **`grouping_config`** selects the new schema and the v3 API; omitting it keeps the previous fields and v2 API. v2-only attributes are **deprecated** with warnings and CHANGELOG links to their replacements.
> 
> The reshuffle: grouping moves to **`grouping_config`**, Slack/Teams + message template merge into **`message_config`**, and the incident template nests under **`incident_config.template`**. **`defer_time_seconds`** / **`auto_relate_grouped_alerts`** become **`escalation_config.when_alert_joins_group`** (mode + optional grace period).
> 
> CRUD **dispatches** on schema mode (plan for create/update, state for read/delete). **Import** probes v3 and falls back to v2 when the org returns `api_not_yet_available`. **ValidateConfig** enforces v2/v3 mutual exclusion and conditional rules at plan time; **mode-aware plan modifiers** avoid perpetual `(known after apply)` on schema-specific computed fields. In-place **v2→v3 migration** is an update, not a replace.
> 
> CHANGELOG, generated docs, and examples are updated to the new layout; v5.41.0 migration notes are in the CHANGELOG.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 03d0b9af90fea2ed870b6fdcacbc81a2fbe9a95c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- `incident_alert_route` now supports the v3 alert routes API in place. Set the new top-level `grouping_config` block to opt into the v3 schema: grouping moves into `grouping_config`, `channel_config` and `message_template` combine into `message_config`, and the incident template nests under `incident_config.template`. The v2-only fields (`channel_config`, `message_template`, `incident_template`, and the `incident_config` grouping fields) are deprecated and emit a warning pointing at their v3 location, but continue to work via the v2 API. You can migrate an existing route in place — add the v3 blocks and drop the deprecated fields in the same change — and Terraform updates the route rather than replacing it. No resource-type change is required.
```
```suggestion
- `incident_alert_route` now supports the v3 alert routes API. When v3's new top-level `grouping_config` block is set, we interpret config via the v3 schema, otherwise we use the v2 schema.
- The following changes are made between v2 and v3:
  - Alert grouping settings are moved out of `incident_config` to a top level `grouping_config`
  - `channel_config` and `message_template` combine into `message_config`
  - Incident template nests under `incident_config.template
- Existing routes can be migrated in place — add the v3 blocks and drop the deprecated fields in the same change — and Terraform updates the route rather than replacing it. Have a look at the "Export" button on the config screen in the dashboard to see the v3 Terraform for your alert routes.
```

internal/provider/incident_alert_route_resource.go
```diff
@@ -444,14 +628,108 @@ func (r *IncidentAlertRouteResource) Delete(ctx context.Context, req resource.De
...
+	claimResource(ctx, r.client, id, &resp.Diagnostics, client.AlertRoute, r.terraformVersion)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	v3Result, err := r.client.AlertRoutesV3ShowWithResponse(ctx, id)
+	switch {
+	case err == nil && v3Result.JSON200 != nil:
+		data := models.AlertRouteResourceModel{}.FromAPIV3(v3Result.JSON200.AlertRoute)
+		resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```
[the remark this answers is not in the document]
acroz: Thanks — real subtlety, but I think this is acceptable as-is rather than needing a fix:

- **It's self-correcting, not a perpetual diff.** Import populates state via `FromAPIV3` (no config/plan context), so an explicit `= []` can come back `null`. But the next `plan`/`apply` runs `FromAPIV3WithPlan` with the config as the plan, which normalises the empty collection back to `[]`, so state converges after a single apply — there's no perpetual update.
- **The acceptance tests already cover it.** The v3 `ImportStateVerify` steps use configs with `destinations = []` and `grouping_keys = []`, and pass in CI, so these round-trip cleanly in practice.
- Import genuinely can't distinguish an omitted optional from an explicit `[]` (there's no config at import time), so the only way to fully remove the one-time post-import diff would be to make these collections `Computed` — a disproportionate change for a self-correcting edge case.

Happy to revisit (e.g. `ImportStateVerifyIgnore`, or making them `Computed`) if we see it bite in practice.

docs/resources/alert_route.md
```diff
@@ -3,16 +3,16 @@
...
   We'd generally recommend building alert routes in our web dashboard https://app.incident.io/~/alerts/configuration, and using the 'Export' flow to generate your Terraform, as it's easier to see what you've configured. You can also make changes to an existing alert route and copy the resulting Terraform without persisting it.
 ---
 
 # incident_alert_route (Resource)
 
-Configure your alert routes in incident.io.
+Alert routes define how alerts are processed: how they're grouped, which channels they post to, who is escalated, and whether they open incidents.
 
-Alert routes define how alerts from different sources are processed, grouped, and routed to the right teams and people.
+This resource supports two configuration schemas for alert routes, corresponding to the v2 and v3 alert route APIs. The principal difference is that in v3, alert grouping configuration is split out of `incident_config` into a top-level `grouping_config`. In addition, `channel_config` and `message_template` are merged together under `message_config`, and the incident template is nested under `incident_config.template`.
```
edds: I find it really awkward that this is exposing the concepts of our 'v2' and 'v3' APIs in these docs. They aren't concepts that someone using this provider should really have to think about. We should describing the new state of the world we want you to use and mark the old attributes as deprecated. We should mention the group of properties that have to be one way or the other and have good changelog describing how to get there.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Unreleased
 
+- `incident_alert_route` now supports a new configuration format for alert routes. Set the new top-level `grouping_config` block to opt a route into it:
+  - Alert grouping moves out of `incident_config` into the top-level `grouping_config`.
+  - `channel_config` and `message_template` are replaced by `message_config`.
+  - The incident template moves under `incident_config.template`.
+- The previous attributes (`channel_config`, `message_template`, `incident_template`, and the grouping fields inside `incident_config`) are now deprecated but continue to work. Migrate a route in place by adding the new blocks and removing the deprecated fields in the same change — Terraform updates the route rather than replacing it. The easiest way to get the new configuration is the "Export" button on the alert route config screen in the dashboard.
```
edds: This feels like you're hiding that these two formats are incompatible with each other, and really burying that the old attributes are now deprecated. It should be clear in the changelog that we've introduced `grouping_config` and `message_config`. And super clear that we've deprecated `channel_config`, `message_template` etc.. and these deprecated fields will be removed in an upcoming major version update. People can look at resource docs to work out how they are used, but unless they realise that they need to based on what they read here they won't look.

internal/provider/incident_alert_route_resource.go
```diff
@@ -42,7 +55,11 @@ func (r *IncidentAlertRouteResource) Metadata(ctx context.Context, req resource.
...
+
+An alert route is configured using one of two mutually-exclusive sets of attributes. The current format configures alert grouping via the top-level ` + "`grouping_config`" + ` block, channels and message under ` + "`message_config`" + `, and the incident template under ` + "`incident_config.template`" + `. Setting ` + "`grouping_config`" + ` selects this format. The deprecated format (used when ` + "`grouping_config`" + ` is omitted) instead configures grouping inside ` + "`incident_config`" + ` and uses the ` + "`channel_config`" + `, ` + "`message_template`" + `, and ` + "`incident_template`" + ` blocks; these attributes are deprecated and should be migrated to their current equivalents. You can't mix attributes from the two sets.
+
+We'd generally recommend building alert routes in our [web dashboard](https://app.incident.io/~/alerts/configuration), and using the 'Export' flow to generate your Terraform, as it's easier to see what you've configured. You can also make changes to an existing alert route and copy the resulting Terraform without persisting it.`,
```
edds, line 67:
```suggestion
		MarkdownDescription: `Alert routes define how alerts are processed: how they're grouped, which channels they post to, who is escalated, and whether they open incidents.
		
We'd generally recommend building alert routes in our [web dashboard](https://app.incident.io/~/alerts/configuration), and using the 'Export' flow to generate your Terraform, as it's easier to see what you've configured. You can also make changes to an existing alert route and copy the resulting Terraform without persisting it.`,

A number of fields are been marked as deprecated (`channel_config`, `message_template`, `incident_template`), if you use the new `grouping_config` attribute none of those fields will continue to work. Deprecated attributes will be removed from the next major version of this provider.
```
In my opinion I'd be simpler here and put the detail in the changelog. The people that care about how to move things are people upgrading and they will look at the changelog for upgrade advice. Anyone else is coming here for the first time and explaining too much about previous formats isn't useful.

internal/provider/incident_alert_route_resource.go
```diff
@@ -143,22 +162,137 @@ func (r *IncidentAlertRouteResource) Schema(ctx context.Context, req resource.Sc
...
+								MarkdownDescription: apischema.Docstring("AlertRouteWhenAlertJoinsGroupV3", "grace_period_seconds"),
+							},
+						},
+					},
+				},
+			},
+			// --- v3-only: grouping_config (the schema discriminator) ---
+			"grouping_config": schema.SingleNestedAttribute{
+				Optional:            true,
+				MarkdownDescription: apischema.Docstring("AlertRouteV3", "grouping_config") + "\n\nSetting this block selects the current configuration format for the alert route.",
```
edds: I feel like Claude has used `current configuration format` or `current format` to get around saying v2/v3 but it reads just as odd. I'd drop the  `+ "\n\nSetting this block selects the current configuration format for the alert route.",`

internal/provider/incident_alert_route_resource.go
```diff
@@ -143,22 +162,137 @@ func (r *IncidentAlertRouteResource) Schema(ctx context.Context, req resource.Sc
...
 						},
 					},
+					// --- v3-only escalation field ---
+					"when_alert_joins_group": schema.SingleNestedAttribute{
+						// Optional + Computed: v3 only. When grouping is enabled the API
+						// returns a default when_alert_joins_group even if the user
+						// didn't configure one, so we accept that computed value.
+						Optional:            true,
+						Computed:            true,
+						MarkdownDescription: apischema.Docstring("AlertRouteEscalationConfigV3", "when_alert_joins_group") + " Only used with `grouping_config`.",
```
edds: Similarly, I'd be tempted to move this to be a plan validation and not put it here, it makes it sound like you've got to use it when you use `grouping_config` but it's optional so that's not the case.

edds (commented):
I've left a couple of comments, this is much better, but I still feel it might want to go slightly further. The change log should be much clearer as to the migration steps for a person upgrading, it's quite light on detail at the moment and that detail is hidden around the resource. Equally the resource description should be more aimed at explaining what current good looks like that focusing on what the previous functionality was. People can easily look at old versions of the docs if they want to know how old providers were set up.

internal/provider/incident_alert_route_resource.go
acroz: > move this to be a plan validation

Are you suggesting to make this not be a computed field? The computed field is really just hear to handle a default returned by the server (hence avoiding conflicts with the local config when this was unset). All the validation happens in `ValidateConfig`, I believe.

edds: Oh no, sorry leave it as a computed field, you've already got the validation, so this was just about avoiding talking about the `grouping_config` in the description. This is all good now.

edds (approved):

Issue #390, opened 2026-07-15
'Provider produced inconsistent result after apply' while creating incident_alert_route resource


cp-rodrigobrasil:
# Description

Provider error while creating `incident_alert_route` resource. The route is created on UI, but Terraform shows this error:

```txt
incident_alert_route.datadog_route_production: Destroying... [id=xxx]
incident_alert_route.datadog_route_production: Destruction complete after 1s
incident_alert_route.datadog_route_production: Creating...

Error: Provider produced inconsistent result after apply

When applying changes to incident_alert_route.datadog_route_production, provider
"provider[\"registry.terraform.io/incident-io/incident\"]" produced an unexpected new value:
.escalation_config.escalation_targets: planned set element
cty.ObjectVal(map[string]cty.Value{"escalation_paths":cty.ObjectVal(map[string]cty.Value{"array_value":cty.ListVal([]cty.Value{cty.ObjectVal(map[string]cty.Value{"literal":cty.NullVal(cty.String),
"reference":cty.StringVal("expressions[\"xxx\"]")})}),
"value":cty.NullVal(cty.Object(map[string]cty.Type{"literal":cty.String, "reference":cty.String}))}),
"users":cty.NullVal(cty.Object(map[string]cty.Type{"array_value":cty.List(cty.Object(map[string]cty.Type{"literal":cty.String,
"reference":cty.String})), "value":cty.Object(map[string]cty.Type{"literal":cty.String,
"reference":cty.String})}))}) does not correlate with any element in actual.

This is a bug in the provider, which should be reported in the provider's own issue tracker.
```

# Configuration

Simple route copied from the UI:

```hcl
resource "incident_alert_route" "datadog_route_production" {
  name       = "Datadog Route Production"
  enabled    = true
  is_private = false

  alert_sources = [
    {
      alert_source_id = incident_alert_source.datadog.id

      # Conditions for including alerts from this source
      condition_groups = []
    },
  ]

  # The conditions that must be met for this alert route to fire on a whole-route basis
  condition_groups = [
    {
      conditions = [
        {
          operation      = "is_set"
          subject        = "alert.attributes.xxx"
          param_bindings = []
        },
      ]
    },
  ]

  # Configuration for how alerts are grouped together on this route
  grouping_config = {
    default = {
      enabled = false
    }
  }

  # Configuration for which users or escalation paths should be notified
  escalation_config = {
    auto_cancel_escalations = true

    escalation_targets = [
      {
[...]

PR #399, opened 2026-07-16
Fix inconsistent result after apply when alert route escalation targets use escalation_paths

grahamsk:
## Problem

Creating (or replacing) an `incident_alert_route` whose `escalation_config.escalation_targets` use `escalation_paths` fails with:

```
.escalation_config.escalation_targets: planned set element ... does not correlate with any element in actual.
```

Reported in #390 (closed, but no fix has shipped as of v5.44.0 — reproduced there with UI-exported config using `reference`, and independently by us on 5.43.0 with `literal` bindings).

## Root cause

The API echoes each escalation target's `escalation_paths` binding back in the target's `users` field, even when the create/update request never set `users` (confirmed from a failed apply's stored state: the escalation path id appears duplicated into `users.array_value`). `FromAPIV2WithPlan` and `FromAPIV3WithPlan` mapped `target.Users` unconditionally when non-nil, so the applied state carried `users` where the plan had null, and the framework rejects the mismatch. The client structs and binding conversions are clean — this is purely the readback of an API-side echo, the same class described in this repo's own guidance about API responses returning nested objects the user didn't configure.

## Fix

Reconcile `users` against the plan (the existing pattern used for the template name binding, custom fields, and v3 grouping), via a shared helper used by both the v2 and v3 conversion paths:

- Drop `users` **only** in the precise echo case: the API target carries `escalation_paths`, the plan contains a target with that same `escalation_paths` binding (matched by content — it's a set), and that plan target did not configure `users`.
- Preserved verbatim: users-only targets, targets that configure both fields (exact match is preferred, so a users-configured target keeps its value even when a sibling target shares the same paths binding), targets with no matching plan entry (out-of-band edits still surface as drift), and the nil-plan import path.

One case is undetectable by construction and documented in a code comment: `users` added out-of-band to the exact target Terraform manages via `escalation_paths` is indistinguishable from the echo.

## Tests

- `TestAlertRouteV2EscalationTargetUsersEcho` (table-driven: echo with `literal`, echo with `reference` — the #390 reporter's shape, users-only + echoed sibling, both-fields-configured, nil-plan import passthrough, unmatched-target drift) and `TestAlertRouteV3EscalationTargetUsersEcho` mirroring the echo and preservation cases through the v3 path.
- The echo cases fail on current master and pass with the fix.
- `go build ./...`, `go vet ./internal/provider/...`, and `go test ./internal/provider/models` all pass. `TestAccIncidentMaintenanceWindowResource` fails identically on unmodified master in my environment (no `INCIDENT_API_KEY`), so acceptance runs are left to your CI.

## Notes

- After this fix, a route **imported** while the API still echoes will show a one-time diff removing the echoed `users`; the subsequent apply converges.
- `incident_maintenance_window_resource.go` maps escalation-target `users` with the same unconditional pattern — if the API echoes there too it has the same latent bug, but I've left it out of scope (no reported issue, and its `buildModel` takes no plan today).
- No generated code touched.

---

> [!NOTE]
> **Medium Risk**
> Changes how `incident_alert_route` escalation target state is built on create/update/read; mis-reconciliation could hide drift or drop legitimate `users`, though scope is narrow and well-tested.
> 
> **Overview**
> Fixes **Provider produced inconsistent result after apply** for `incident_alert_route` when `escalation_config.escalation_targets` use **`escalation_paths`** without **`users`**. The API duplicates each target’s `escalation_paths` binding into `users`; the provider used to persist that echo, so post-apply state no longer matched the plan on the `escalation_targets` set.
> 
> A shared **`escalationTargetFromBindings`** helper (v2 and v3 `FromAPI*WithPlan`) reconciles `users` against the plan: it clears `users` only when the plan has the same `escalation_paths` binding and did not set `users`. Plan matching uses new **`paramBindingsEqual`** helpers on engine param bindings (set elements, not list order). Real `users`, dual-field targets, import with nil plan, and unmatched API targets are left unchanged so drift still shows.
> 
> Unit tests cover echo (literal/reference), mixed targets, import passthrough, and unmatched drift on v2; v3 mirrors the main cases. **CHANGELOG** documents the fix (#390).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f32948a804966871cb2b2eea1e925e8b4c238ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


benfdking:
Thanks for this, and sorry for the slow reply. You got to the right root cause independently, and the write-up (especially pinning it on the set-element correlation rather than the bindings conversion) was spot on. Unfortunately, this has been overtaken by events: #400 landed a fix for the same bug and shipped in **v5.44.1**. 

If you're still on a version before v5.44.1, upgrading should resolve the issue. Note that state written by an older provider carries the echoed `users`, so you'll see a single diff removing it, after which it's stable. 

Thanks again for taking the time to dig in and write it up properly. It's really appreciated.

PR #409, opened 2026-07-27
Make incident_schedule_sync_rule and incident_schedule_sync_target importable

benfdking:
`incident_schedule_sync_rule` and `incident_schedule_sync_target` were the only two resources whose registry pages carried no Import section, so they read as un-importable even though both implement `ResourceWithImportState`. The gap was `examples/resources/<name>/import.sh`: `tfplugindocs` renders that script into the resource page's "Import" section, and without it the section is omitted entirely.

This adds the two missing examples, regenerates the docs, and hardens both `ImportState` implementations so the documented behaviour actually holds. All 18 resource pages now document import.

## Changes

**Import examples and docs** — `import.sh` for both resources, plus the regenerated `docs/resources/schedule_sync_{rule,target}.md`. Targets import by their ID; rules are nested under a schedule in the API, so they import by `schedule_id:rule_id` (the same shape as `incident_catalog_type_attribute`). The target example calls out that an imported target should reference its existing Slack user group with `slack_user_group_id`, because `new_slack_user_group` asks us to *create* a group and leaving it set plans a replacement of the target.

**Full state on import** — both resources previously wrote only their ID attributes and left the rest to the follow-up `Read`. That meant an ID that didn't exist was claimed as a Terraform-managed resource and imported as an empty object before failing later, and the state written during import didn't match what `Read` produces (the class of mismatch behind the `import {}` "Value Conversion Error" fixed for `incident_catalog_type_attribute` in v5.43.1). Both now fetch the resource first, claim it only once it is confirmed to exist, and set the whole model, with a clear diagnostic naming the missing target/rule.

**Import ID validation** — a sync rule import ID with an empty schedule or rule ID (`:rule_id`, `schedule_id:`) is now rejected with the format hint instead of being treated as valid, and the error quotes the ID received. The parsing is extracted into `parseScheduleSyncRuleImportID` with unit test coverage.

**Nil response bodies** — `Read` and `ImportState` on both resources returned a nil-pointer dereference if the API returned an unexpected 2xx without a JSON body; they now return a diagnostic with the status, matching the `incident_escalation_path` fix in v5.41.1.

**DEVELOPING.md** — a short "Import support" section recording that every resource needs an `import.sh` and that `ImportState` should populate the whole model, so this doesn't regress with the next resource.

## Testing

Acceptance tests for these resources need Slack `usergroups:write` and are gated behind `TF_ACC_SLACK_USER_GROUPS`, so they don't run in CI. To verify end to end I ran the built provider (via `dev_overrides`) against a local stub of the three endpoints involved (`ScheduleSyncTargetsV2Show`, `SchedulesV2ShowScheduleSyncRule`, `ManagedResourcesV2CreateManagedResource`):

- `terraform import` of both resources succeeds and writes complete state (`add_bot_to_group`, `slack_user_group_id`, `slack_team_id` for the target; `schedule_id`, `schedule_sync_target_id`, `sync_type`, `rotation_id` for the rule), and the subsequent `terraform plan` reports **No changes**.
- The same via `import {}` blocks plans **2 to import, 0 to add, 0 to change, 0 to destroy** — no drift, no conversion errors.
- Error paths: a non-existent target ID gives `Schedule Sync Target Not Found`; `no-colon` and `:rule_id` give `The import ID must be in the format: schedule_id:rule_id (got …)`; a well-formed ID for a rule that doesn't exist gives `Schedule Sync Rule Not Found`.
- Confirmed the documented `new_slack_user_group` caveat: keeping it in the configuration after import plans a destroy-and-create, which Terraform surfaces itself with "Warning: this will destroy the imported resource".

Also: `go build ./...`, `go generate ./...` (idempotent — only the two doc pages change), unit tests, and `golangci-lint run` (0 issues).

<div><a href="https://cursor.com/agents/bc-0e73a31d-8eec-489a-9f59-0871324c0ce6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-0e73a31d-8eec-489a-9f59-0871324c0ce6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+- Document import support for `incident_schedule_sync_target` and `incident_schedule_sync_rule`, the last two resources whose registry pages had no import example. Targets import by their ID; rules are nested under a schedule in the API, so they import by `schedule_id:rule_id`.
+- Make importing either resource read it from the API up front, so the imported state is fully populated and an ID that doesn't exist fails with a clear message rather than silently importing an empty resource. An import ID with an empty schedule or rule ID is now rejected too, and an unexpected successful API response without a body returns a diagnostic instead of panicking.
+
 ## v5.45.0
 
 - Add `group_alerts_summary` to the v3 `incident_alert_route` resource, on Slack and MS Teams channel targets under `message_config.destinations`. When enabled, grouped alerts render as a single editable group-summary message per channel instead of one message per alert.
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -154,6 +154,22 @@ TF_CLI_CONFIG_FILE=./dev.tfrc terraform plan
 TF_CLI_CONFIG_FILE=./dev.tfrc terraform apply
 ```
 
+## Import support
+
+Every resource supports `terraform import`, and each resource's registry page documents how.
+When you add a resource, implement `resource.ResourceWithImportState` and add an
+`examples/resources/<resource_name>/import.sh` showing the command: `tfplugindocs` renders that
+script into the "Import" section of the resource's page, so without it the resource looks
+un-importable even though the code supports it.
+
+`ImportState` should populate the whole model from the API rather than writing only the ID
+attributes. That keeps imported state identical to what `Read` produces, makes an ID that doesn't
+exist fail with a clear diagnostic instead of importing an empty resource, and avoids the untyped
+collection attributes that break `import {}` blocks with a "Value Conversion Error". For a resource
+nested under a parent (like a sync rule under a schedule), take a composite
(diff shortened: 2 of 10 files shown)


eespinosa (approved):

benfdking:
@cursor can you rebase this foir me please

PR #439, opened 2026-08-03
Add incident_schedule_beta and incident_schedule_rotation_beta

eespinosa:
Adds beta resources and data sources for the v3 schedules API: `incident_schedule_beta` and `incident_schedule_rotation_beta`, each as both a resource and a data source.

A schedule and its rotations are now separate resources, so editing one rotation no longer rewrites the whole schedule. `incident_schedule_beta` covers name, timezone, owning teams and public holidays; `incident_schedule_rotation_beta` covers the users, handover cadence, concurrent shifts and optional working hours.

The existing `incident_schedule` is untouched, still on v2, and not deprecated. The `_beta` suffix is because these schemas may still change incompatibly.

**API schema refresh.** The client had none of the six rotation operations, so the embedded schema is refreshed per DEVELOPING.md. Additive only — 5 paths, 27 schemas, no removals, no client symbols dropped — and `docs/` regenerates with no change to any existing page. Also adds `terraform-plugin-framework-timetypes`.

**For review:**
- `enumValues` is new in `helpers.go`, reading allowed values from the API schema so validation can't fall behind the API.
- Both `UseStateForUnknown` uses are top-level `id` attributes, so the nested-collection fix from #435 doesn't apply here.
- Rebased onto #435; suite passes against framework 1.19.0.

`go build`, `go vet`, `golangci-lint run ./...` (0 issues) and `go test ./...` pass. Acceptance tests not run against a live org.

(Replaces #438, which GitHub closed when I renamed the branch.)

---

> [!NOTE]
> **High Risk**
> New resources drive on-call scheduling via rollout and preview APIs; misconfiguration or stale plans could change paging coverage, though they are additive beta alongside unchanged v2 `incident_schedule`.
> 
> **Overview**
> Introduces **beta** Terraform support for incident.io’s **v3 schedules API**, splitting on-call configuration into **`incident_schedule_beta`** (schedule metadata) and **`incident_schedule_rotation_beta`** (per-rotation line-up and cadence), each with matching **data sources**, **docs**, and **examples**. Legacy **`incident_schedule`** stays on v2 and is unchanged.
> 
> The rotation resource adds v3-only controls such as **`rollout`**, **`scheduling_mode`**, and **`working_intervals`**, and wires **plan/apply** to **`PreviewRotationRollout`** (pinned **`from`**, optional **`expected_effective_from`**) so phased edits don’t silently change who is on call. Plan-time checks block invalid combos (e.g. **`rollout`** with **`first_interval_starts_at`** in one apply, empty **`working_intervals`**).
> 
> Provider plumbing includes a refreshed embedded **OpenAPI** client (v3 rotation CRUD + preview paths), **`enumValues`** in **`helpers.go`** for schema-driven enum validation, and **`terraform-plugin-framework-timetypes`** for handover timestamps. **`CHANGELOG.md`** documents the beta surface and migration/import notes (`schedule_id:rotation_id` for rotations).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc6b472295ac553cb15d00812538fe8abafb9319. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/resources/schedule_beta.md
isaacseymour, line 1: This should have something that explains how it's different to non-beta schedule resource, that v7.0 this will become non-beta and the old resource will be removed, and (maybe?) how to migrate

docs/resources/schedule_rotation_beta.md
```diff
@@ -0,0 +1,141 @@
...
+  ]
+}
+```
+
+<!-- schema generated by tfplugindocs -->
+## Schema
+
+### Required
+
+- `concurrent_shifts` (Number) How many shifts run at the same time, which is how many people are on call at once. Reducing this stops scheduling the last of them, and stops any overrides on those shifts from applying.
```
isaacseymour: I kinda think this should be optional and default to 1!

docs/resources/schedule_rotation_beta.md
```diff
@@ -0,0 +1,141 @@
...
+- `handovers` (Attributes List) The cadence shifts hand over on, applied in turn (see [below for nested schema](#nestedatt--handovers))
+- `name` (String) Human readable name for the rotation, unique within the schedule
+- `schedule_id` (String) ID of the schedule this rotation belongs to. Changing this replaces the rotation, as a rotation can't move between schedules.
+- `users` (List of String) IDs of the people in the rotation, in the order they take shifts.
+
+### Optional
+
+- `rank` (Number) Where this rotation sits in the schedule's running order, lowest first. Unset when it has never been ordered.
+- `rollout` (String) How a change to this rotation is introduced: `immediate` replaces the line-up now and can change who is on call this minute, `after_current_shift` lets the shift on call finish first, and `after_full_rotation` waits for everyone to have taken a turn. Leave it out to replace the line-up straight away. Creating a rotation ignores it, as there's no shift to protect yet.
+- `scheduling_mode` (String) How users are allocated across shifts of differing length. One of fair, sequential. Omit it to leave us to pick.
```
isaacseymour: I think we should explain the difference between fair and sequential, and when this is relevant. I _think_ it's only relevant if you have handovers of differing lengths (e.g. "5d / 2d" handovers, or "3d/4d" handovers)?

docs/resources/schedule_rotation_beta.md
```diff
@@ -0,0 +1,141 @@
...
+- `working_intervals` (Attributes List) If set, restricts on-call to these weekday intervals. Omit it to keep the rotation on call around the clock — an empty list isn't a way to say that. (see [below for nested schema](#nestedatt--working_intervals))
```
isaacseymour: Is an empty list valid? If not I think we should just say "An empty list is not valid."

internal/apischema/public-schema-v3-including-secret-endpoints.json
isaacseymour, line 1: let's mark this as generated in `.gitattributes`!

internal/provider/incident_schedule_rotation_beta_resource.go
```diff
@@ -0,0 +1,1321 @@
...
+
+func (r *IncidentScheduleRotationBetaResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: `Manage a rotation on an on-call schedule.
+
+A rotation is a group of people who take turns being on call, and the cadence they
+hand over on. A schedule can hold several, and each is managed independently — so
+changing one never rewrites the others.
+
+Editing a rotation takes effect straight away.`,
```
isaacseymour: Isn't this dependent on the `rollout` attr?

eespinosa: It is! It's being updated right now 👀

internal/provider/incident_schedule_rotation_beta_resource.go
```diff
@@ -0,0 +1,1321 @@
...
+	resp.Diagnostics.Append(diags...)
+	if diags.HasError() || users.IsNull() || users.IsUnknown() {
+		return
+	}
+
+	if len(users.Elements()) == 0 {
+		resp.Diagnostics.AddAttributeError(
+			usersPath,
+			"Empty users",
+			"A rotation needs at least one person to put on call. Remove the rotation instead if it shouldn't schedule anyone.",
```
isaacseymour: Should we mention that you can use the special user ID `NOBODY` to create a rotation that doesn't actually schedule anyone unless you whack an override in...? This is how people set up (e.g.) shadow rotations!

internal/provider/incident_schedule_rotation_beta_resource.go
```diff
@@ -0,0 +1,1321 @@
...
+			"Unknown scheduling_mode",
+			fmt.Sprintf("Set one of %s, or remove the attribute to leave us to pick.",
+				strings.Join(scheduleRotationSchedulingModes, ", ")),
+		)
+	}
+}
+
+// knownString returns the value of a string attribute, and false when it's missing, null
+// or unknown — none of which can be judged at plan time.
+func knownString(value attr.Value) (string, bool) {
```
isaacseymour: feels like these should go in the `helpers.go` and be reused in like...loads of resources (soon)

internal/provider/incident_schedule_rotation_beta_resource.go
```diff
@@ -0,0 +1,1321 @@
...
+	// exactly as planned — so left alone this surfaces as an apply that blames the
+	// provider for a bug. Say what's actually happening, before anything is written.
+	timezone := r.scheduleTimezone(ctx, state.ScheduleID.ValueString())
+	if !anchorToState(previewed, plan.FirstIntervalStartsAt, timezone).Equal(plan.FirstIntervalStartsAt) {
+		resp.Diagnostics.AddAttributeError(
+			path.Root("first_interval_starts_at"),
+			"Handover time can't move while a change is phased in",
+			"This rotation hands over on the cadence it runs today, and that's what the new "+
+				"line-up is anchored to, so this handover time would be dropped. Move it in an "+
+				"apply of its own — or one without rollout set — then phase the rest in.",
```
isaacseymour: I'm finding this a bit confusing. What should the user actually do here?

internal/provider/incident_schedule_rotation_beta_resource.go
```diff
@@ -0,0 +1,1321 @@
...
+		names = append(names, fmt.Sprintf("%q", lo.FromPtr(layer.Name)))
+	}
+
+	shiftsPath := path.Root("concurrent_shifts")
+	r.warnLostOverrides(ctx, resp, state, droppedIDs, &shiftsPath, func(count string) string {
+		return fmt.Sprintf("Reducing concurrent shifts from %d to %d on rotation %q removes "+
+			"on-call layer(s) %s, which stops %s targeting them from applying. The overrides "+
+			"aren't deleted, but nothing renders one once its layer is gone, and adding the "+
+			"shift back creates a new layer rather than restoring this one. Recreate the "+
+			"overrides you still need on a remaining shift.",
```
isaacseymour: I think this is a bit hard to understand. How about something like:
> Reducing concurrent shifts from %d to %d on rotation %q means that some overrides will no longer have a shift to be applied to, and will no longer have any effect on the schedule. You may wish to review overrides to make sure the right people are on-call after this change.

?

isaacseymour (approved):

PR #452, opened 2026-08-11
Validate alert source templates at plan time

eespinosa:
An `incident_alert_source`'s expressions were only checked when you applied. A reference that didn't resolve, or a merge strategy the attribute doesn't support, failed part way through the apply — after Terraform had already created other resources.

`ModifyPlan` now asks the API whether the planned template would be accepted, using `POST /v2/alert_sources/actions/validate`. A rejection fails the plan and repeats what the API said, field path included, so the plan-time error reads like the apply-time one.

It isn't in `ValidateConfig`, which `terraform validate` also calls: that runs the provider without configuring it, so there's no client to ask with.

## What it deliberately doesn't check

**Templates Terraform hasn't settled.** An expression pointing at a catalog type the same apply creates is unknown until it exists, and checking around the gaps would report errors the apply won't hit.

Optional+Computed attributes are exempt from that, which matters more than it sounds: `merge_strategy` is unknown for every binding whose config doesn't set one, so counting it would have skipped the check on most configs that have a binding at all. The API fills it in either way, and the payload leaves it out. There's a test for this specifically.

**Sources the plan doesn't change.** There's no apply to fail, and checking anyway would cost a request per alert source on every plan.

**Anything that isn't a rejection warns rather than fails.** A 404 before the endpoint is available, a 500, a timeout — the check didn't run, which doesn't mean the config is wrong. The call is bounded by a short timeout for the same reason: the shared client retries a 5xx ten times with backoff, which would otherwise stall a plan for minutes per alert source over an advisory check.

## Testing

Unit tests cover a valid template, a rejection, an unavailable check, a template with unknown values, an unchanged source, a destroy, and the computed-values case above. The last one was verified to fail against the earlier, coarser gate.

Also driven end to end against the real endpoint, with this provider wired in via `dev_overrides`:

- a template binding a non-existent attribute fails the plan with the API's own message and `"pointer": "template.attributes"`
- the same config with that attribute removed plans cleanly

## Notes for review

- The vendored client is a wholesale refresh, so `client.gen.go` and the schema carry unrelated API surface that had accumulated since they were last regenerated. The hand-written change is `incident_alert_source_resource.go` plus the test.
- `owningTeamIDsPayload` is extracted because plan and apply have to send the same team IDs — the permission check depends on them. Unknown elements are dropped, so a plan doesn't send `""` as a team ID.
- Merging should wait until the endpoint is available, or every plan warns until it is.

### DIFF
--- a/internal/provider/incident_alert_source_resource.go
+++ b/internal/provider/incident_alert_source_resource.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/http"
+	"time"
 
 	"github.com/hashicorp/terraform-plugin-framework/diag"
 	"github.com/hashicorp/terraform-plugin-framework/path"
@@ -14,6 +16,7 @@ import (
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
 	"github.com/hashicorp/terraform-plugin-framework/types"
+	"github.com/hashicorp/terraform-plugin-go/tftypes"
 	"github.com/hashicorp/terraform-plugin-log/tflog"
 
 	"github.com/incident-io/terraform-provider-incident/internal/apischema"
@@ -25,6 +28,7 @@ var (
 	_ resource.ResourceWithConfigure      = &IncidentAlertSourceResource{}
 	_ resource.ResourceWithImportState    = &IncidentAlertSourceResource{}
 	_ resource.ResourceWithValidateConfig = &IncidentAlertSourceResource{}
+	_ resource.ResourceWithModifyPlan     = &IncidentAlertSourceResource{}
 )
 
 type IncidentAlertSourceResource struct {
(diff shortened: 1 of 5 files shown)


internal/provider/incident_alert_source_resource.go
```diff
@@ -252,6 +256,143 @@ func NewIncidentAlertSourceResource() resource.Resource {
...
+// slow-but-working API still answers, short enough that an unhealthy one costs a plan
+// seconds rather than minutes. A var so tests don't have to wait it out.
+var alertSourceValidateTimeout = 10 * time.Second
+
+// ModifyPlan asks the API whether the planned template would be accepted, so a broken
+// expression shows up in the plan instead of part way through an apply.
+//
+// It can't live in ValidateConfig, which is also called by `terraform validate` — that
+// runs the provider without configuring it, so there's no client to ask with.
+func (r *IncidentAlertSourceResource) ModifyPlan(ctx context.Context, req resource.ModifyPlanRequest, resp *resource.ModifyPlanResponse) {
```
yuhanwwu, line 269:
```suggestion
if req.Plan.Raw.Equal(req.State.Raw) {
    return
}
```
so we skip unchanged sources

yuhanwwu (approved):

PR #373, opened 2026-07-01
Add support for workflow form fields

paprikati:
## What

Adds a `form_fields` attribute to the `incident_workflow` resource, so you can define the fields shown to a user when they manually trigger a workflow. The value a user provides becomes available in the workflow scope under `workflow_form.<key>` for use by conditions and steps.

Each form field supports:
- `key` (required) – the key used to reference the field in the workflow scope
- `title` (required) – human readable title shown in the form
- `type` (required) – the engine resource type of the field (e.g. `Text`, `User`, `IncidentSeverity`)
- `array` (required) – whether the field holds a list of values
- `required` (required) – whether the field must be filled in when running the workflow
- `description` (optional) – help text shown beneath the field
- `id` (optional/computed) – stable identifier, generated if omitted

```hcl
resource "incident_workflow" "page_execs" {
  name    = "Page execs"
  trigger = "manual"
  # ...
  form_fields = [
    {
      key         = "reason"
      title       = "Reason for paging"
      type        = "Text"
      description = "Why are we paging the execs?"
      array       = false
      required    = true
    },
  ]
}
```

## How

The `WorkflowFormFieldV2` / `WorkflowFormFieldPayloadV2` schema and the `form_fields` property on the workflow schemas now come from `core` (they landed in the schema refresh on `master`), so this PR is purely the provider-side wiring against the generated client:

- Added the `form_fields` block to the resource schema and an `IncidentWorkflowFormField` model.
- Wired it into the create/update payloads (`toPayloadFormFields`) and the model builder (`buildFormFields`).
- `id` is optional/computed with a `UseStateForUnknown` plan modifier so omitted ids stay stable across plans and update payloads keep the same field identity.
- `buildFormFields` returns `nil` for a nil **or empty** API response, so workflows without form fields don't show a perpetual diff.
- Added roundtrip unit tests, an example, regenerated docs, and a changelog entry.

## Testing

- `go build ./...`, `go vet ./...` and `gofmt` are clean.
- `TestWorkflowFormFieldsRoundTrip` / `TestWorkflowFormFieldsNilRoundTrip` pass (`go test ./internal/provider/ -run TestWorkflowFormFields`).
- Docs regenerated with `tfplugindocs` (only `docs/resources/workflow.md` changes).
- Acceptance tests require a real API/credentials, so they weren't run here.

[...]

PR #480, opened 2026-08-26
Add support for workflow form fields

benfdking:
Replaces #373, rebased onto `master` with all the Bugbot review comments addressed. Original work by @paprikati; that PR can be closed in favour of this one.

## What

Adds a `form_fields` attribute to `incident_workflow`, so you can define the fields shown to a user when they manually trigger a workflow. The value a user provides becomes available in the workflow scope under `workflow_form.<key>` for use by conditions and steps.

```hcl
resource "incident_workflow" "page_execs" {
  name    = "Page execs"
  trigger = "manual"
  # ...
  form_fields = [
    {
      key         = "reason"
      title       = "Reason for paging"
      type        = "Text"
      description = "Why are we paging the execs?"
      array       = false
      required    = true
    },
  ]
}
```

Each field has a `key`, `title`, `type`, `array` and `required` setting, plus an optional `description` and a computed `id`.

## Rebase

The `WorkflowFormFieldV2` / `WorkflowFormFieldPayloadV2` schemas and the `form_fields` property landed on `master` in the schema refresh, so this is purely provider-side wiring against the generated client — no schema or client changes. That also resolves the **Duplicate form field OpenAPI schemas** finding on #373: `master` now has a single definition of each, on the `key`/`title` shape the provider uses.

The one real conflict was `incident_workflow_resource.go`, where `master` changed `buildModel` to take a `prior` model. That turned out to be a better home for the form field reconciliation than the three call sites the old branch patched, so it now hangs off `prior` alongside `ReconcileOperations`.

## Review comments addressed

**Unknown form field IDs sent empty** (high) — `toPayloadFormFields` called `ValueStringPointer()` on the `id`, which returns a pointer to `""` for an unknown value, and `omitempty` still encodes that as `"id": ""`. Every field on create hit this. Now null and unknown ids are left out of the payload entirely, so the API assigns one. Regression-tested at the JSON level, and I confirmed the test fails against the old code.

**Omit `form_fields` fails to clear** (high) — the payload omitted the key when the attribute was unset, so the API kept the existing fields, and the read-back then adopted them into state against a plan of null: `Provider produced inconsistent result after apply`.

Bugbot suggested resolving this by documenting omit as "leave unchanged", following `permanent_member_user_ids` on `incident_schedule_sync_rule`. I went the other way and made the code match what the changelog already promised — omit clears — because `form_fields` is part of the workflow's own definition, like `steps` and `condition_groups`, rather than a reference to external objects the way permanent members are. Config is the source of truth: the payload now always carries the list. That also avoids leaving fields stranded outside Terraform's control with no well-defined way to re-adopt them. Happy to flip it if you'd rather match the sync-rule contract.

**Plan modifier ignores unknown IDs** (low) — an `id` set from a variable or resource reference could be overwritten by key correlation. Rather than patch around it, `id` is now `Computed` only, not `Optional+Computed`: there's no precedent in this provider for a user-settable server-generated nested id, users have no id to supply for a new field, and allowing one hands them a way to point two fields at the same one. The guard is still there in case that ever changes, and a test pins the attribute to `Computed`.

**Empty API form fields not nil** and **Form field id lacks plan modifier** were already fixed on the branch and are carried over. The `UseStateForUnknown` from the latter is still replaced by `formFieldIDPlanModifier`, since it correlates by list position and `form_fields` has to stay ordered.

## Also

- `ValidateConfig` now rejects two form fields sharing a key. It's meaningless for `workflow_form.<key>`, and it's the assumption the id correlation rests on. The two checks judge their own diagnostics, so a duplicate key doesn't mask an unrelated error elsewhere in the config.
- Added unit tests driving `formFieldIDPlanModifier` the way the framework does (positional prefill and all), which covers the delete/insert/shift cases that previously only had the pure helper tested. Verified they fail if the modifier is neutered.

## Testing

`go build`, `go vet`, `gofmt` and `go generate` are clean, and the full unit suite passes.

Acceptance tests need real credentials, so they didn't run here — worth a run before merge. Two things they'd confirm that I couldn't:

- that the API accepts `"form_fields": []` as "clear these", which the omit-clears behaviour depends on
- that an unset `description` comes back absent rather than as `""`. `buildFormFields` maps it faithfully via `StringPointerValue`, so if core serializes an empty string the apply would fail as an inconsistent result. `TestAccIncidentWorkflowResourceFormFields` asserts this.

---

> [!NOTE]
> **Medium Risk**
> Changes workflow create/update payloads and state reconciliation for an ordered nested list with computed IDs; mistakes could corrupt form field identity on reorder/delete, though coverage is extensive.
> 
> **Overview**
> Adds optional **`form_fields`** to **`incident_workflow`** so manually triggered workflows can define an ordered form; submitted values are exposed in scope as **`workflow_form.<key>`** for steps and conditions. Each field supports **`key`**, **`title`**, **`type`**, **`array`**, **`required`**, optional **`description`**, and a computed **`id`**.
> 
> Provider behavior treats config as source of truth: create/update always send **`form_fields`** (including **`[]`** when unset or omitted) so existing fields are cleared instead of left outside Terraform. Read path maps empty API lists to null unless the user planned **`form_fields = []`**, via **`reconcileFormFields`**. **`formFieldIDPlanModifier`** keeps stable **`id`**s keyed by field **`key`** when the list is reordered or fields are removed, and plan-time validation rejects duplicate keys. Docs, examples, and changelog are updated; unit and acceptance tests cover round-trip, clearing, payload encoding, and id correlation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1095afc82017d064c2ecdf820b0ee3c202adca76. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


benfdking:
Superseded by #481, which is the same single commit on a clean branch.

PR #528, opened 2026-09-02
Add the incident_policy resource and data source

eespinosa:
Adds `incident_policy` as a resource and a data source.

A policy carries exactly one config block — `follow_up`, `debrief`, `post_mortem`, `schedule`, `on_call_readiness` or `vacation_conflict` — and which one it is makes it that type, so `policy_type` is computed rather than written. Swapping which block is set replaces the policy; editing one in place is an ordinary update.

The first commit refreshes the vendored OpenAPI schema and regenerates the client, which is where the policies endpoints come from. Same generator version, so it's a schema change rather than a rewrite — kept separate to keep it out of the reviewable diff.

Draft because the acceptance tests haven't been run against a real API yet.

---

> [!NOTE]
> **Medium Risk**
> New resource drives live policy enforcement (reminders, on-call readiness, schedule gaps); misconfiguration or swapping policy types triggers replacement and can change org behavior. Scalar/array assignee reconciliation is policy-specific but well-tested.
> 
> **Overview**
> Adds **`incident_policy`** so Terraform can create and update incident.io **policies** (scope via `condition_groups`, requirements, due dates, assignees/reminders). There is no writable `policy_type`: you set **exactly one** of `follow_up`, `debrief`, `post_mortem`, `schedule`, `on_call_readiness`, or `vacation_conflict`, and `policy_type` is **computed** from that block. **Changing which block is set replaces the policy**; editing the same block is a normal update. `on_call_readiness` and `vacation_conflict` always assign the violator, so **`assignment_rules` is rejected** alongside them (and API-invented rules are dropped on read for those types).
> 
> Adds **`incident_policy` data source** to look up dashboard-owned policies by **`id` or `name`** (exactly one; plan-time validation), exposing `name`, `description`, `status`, and `policy_type` without the full config.
> 
> Provider support includes **import by ID**, shared engine pieces (`expressions`, param-binding shorthands), and **`ReconcileScalarBinding`** so assignee bindings written as scalars (`value_literal`) do not fail apply when the API returns a one-element `array_value`. **CHANGELOG**, registry docs, and examples cover post-mortem, on-call readiness, and vacation-conflict shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7866e4b9f32b1539bbf04d54cc61eb6a7be2db8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/resources/policy.md
```diff
@@ -0,0 +1,927 @@
...
+so carries no block.
+
+## Example Usage
+
+```terraform
+# The person to chase when a post-mortem falls due. Assignees can also be a
+# reference, such as the incident lead; see `assignment_rules.bindings`.
+data "incident_user" "postmortem_owner" {
+  email = "quality@example.com"
+}
```
benfdking: I think it would be worth splittinig each of these intop their own example.

benfdking (approved):
Small thing

docs/resources/policy.md
eespinosa: Done — split into one example per policy type, with a template that titles each section like the workflow docs do.

PR #556, opened 2026-09-07
Add pausing and resuming heartbeat alert sources

benfdking:
Adds `disabled` to `incident_alert_source` and `incident_alert_source_beta` so a heartbeat can be paused and resumed from Terraform.

Set `disabled = true` to pause monitoring without deleting the source, and `disabled = false` to resume it. Only heartbeat sources support this; any other `source_type` is rejected at plan time. Leaving the attribute unset keeps whatever the source is currently doing, so a pause made in the dashboard survives an unrelated apply.

The create payload has no `disabled` field, so a source that starts paused is created and then immediately updated. On V3 (`incident_alert_source_beta`) the API returns `disabled`, so drift is visible. On V2 the show payload doesn't include it, so the provider keeps the planned value.

## The two APIs disagree about pausing a heartbeat that has never pinged

The v3 endpoint refuses it:

```
status 422: This alert source cannot be disabled until it has received its first event
```

The v2 endpoint accepts the same request on the same kind of freshly-created source and returns 200. Both are exercised by the acceptance tests on this branch, so this isn't inference — `TestAccAlertSourceBetaHeartbeatDisabled` asserts the refusal and `TestAccAlertSourceResource_HeartbeatDisabled` pauses and resumes successfully.

Worth someone who knows the core validation confirming which behaviour is intended. Since v2 doesn't return `disabled`, there's no way from the provider to tell whether its 200 actually paused the source or accepted the field and ignored it. The schema documents the restriction only on the beta resource, where the API enforces it.

Nothing in the provider hardcodes the rule: a paused create still attempts the pause and reports what the API says. When that's refused, the source that was already created stays in state, so the next apply manages it rather than creating a second heartbeat that monitors with nobody looking after it.

## Example

```hcl
resource "incident_alert_source_beta" "nightly_backup" {
  name        = "Nightly backup"
  source_type = "heartbeat"

  heartbeat_options = {
    interval_seconds = 86400
  }

  # Once the backup job has pinged this source at least once.
  disabled = true
}
```

## CI

`TestAccIncidentAPIKeyResourceTeamPairing` fails on this PR's checks and has nothing to do with these changes: it arrives from master via #559, which never had a green run of its own (its push build was cancelled). The step expects `it needs team_ids saying which`, and the diagnostic Terraform renders wraps that sentence, so the pattern can't match:

```
team_role_names grants roles for particular teams, so it needs
team_ids saying which.
```

Left for that PR's author to fix on master rather than patched here.

<div><a href="https://cursor.com/agents/bc-d870a6b7-282c-4769-83a8-73bd44952492?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d870a6b7-282c-4769-83a8-73bd44952492&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


eespinosa (approved):

benfdking:
@curosr can you rebase this all for me please

PR #557, opened 2026-09-07
Add incident_secret resource and data source

eespinosa:
Manages secrets: named credentials a workflow can reference, such as the auth token for an outgoing webhook.

A secret's value is write-only — it's stored encrypted and never returned by the API — so the resource takes it as `value_wo`, a [write-only attribute](https://developer.hashicorp.com/terraform/language/resources/ephemeral/write-only). Terraform can't see a write-only value change, so `value_wo_version` is what asks for a rotation:

```hcl
resource "incident_secret" "pagerduty_webhook_token" {
  name = "PagerDuty webhook token"

  value_wo         = var.pagerduty_webhook_token
  value_wo_version = 1 # change this to rotate
}
```

Both value attributes are optional once the secret exists, so a config can manage a secret's name, description and owning teams while something else looks after its value.

Also adds the `incident_secret` data source, for referencing a secret by `id` or `name`.

Setting a value needs Terraform 1.11 or OpenTofu 1.11 and above.

---

> [!NOTE]
> **Medium Risk**
> Introduces credential create/rotate/delete via API and write-only rotation semantics where misconfiguration (e.g. import plus value_wo_version) can rotate production secrets; values are not persisted in state but still cross the provider at apply time.
> 
> **Overview**
> Adds **`incident_secret`** and **`incident_secret`** data source so workflows can use named credentials from Terraform. The resource sets the plaintext via **`value_wo`** (write-only; Terraform 1.11+ / OpenTofu 1.11+), never stores it in state, and uses **`value_wo_version`** as the visible signal to rotate when **`value_wo`** changes. Metadata (name, description, **`owning_team_ids`**) can be managed without owning the value once the secret exists; import omits **`value_wo_version`** unless you intend a rotation on first apply.
> 
> The data source resolves by **`id`** or **`name`** (exactly one; plan-time validation), with name lookup paging the list API. Updates always send metadata (including empty description / team list when cleared). **`Makefile`** sets **`testacc`** to **`-timeout 18m`** so the growing acceptance suite does not hit Go’s default 10m limit.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b98bc65f63134bc2f97495d5403d666c11067c2e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


benfdking (approved):

docs/resources/secret.md
```diff
@@ -0,0 +1,128 @@
...
+variable "pagerduty_webhook_token" {
+  type      = string
+  ephemeral = true
+}
+
+resource "incident_secret" "pagerduty_webhook_token" {
+  name        = "PagerDuty webhook token"
+  description = "Auth token for the PagerDuty outgoing webhook"
+
+  value_wo         = var.pagerduty_webhook_token
```
benfdking, line 58: I wonder if it is worth adding an example where we use a token in a workflow just to show it off.

PR #560, opened 2026-09-08
Show how to use a secret in a workflow's webhook step, and cap its name at plan time

eespinosa:
Two changes to `incident_secret`.

**The docs never showed how to use a secret**, and the way one reaches a `webhook.send` step isn't guessable. The step takes one two different ways, and the new example uses both:

```hcl
# Substituted when the workflow runs, because the header's feature set allows secrets
data "incident_rich_text" "authorization_header" {
  markdown    = "Authorization: Bearer {{ ${incident_secret.partner_webhook_token.id} }}"
  feature_set = "plain_single_line_with_secrets"
}

# ...while the signing secret takes the ID directly, to sign the request rather than send it
{ value_literal = incident_secret.partner_webhook_token.id }
```

Both reference the secret's `id`, never its value. Rendering it needs a `templates/resources/secret.md.tmpl`, since tfplugindocs only picks up `resource.tf` without one.

I applied the example against a local instance rather than trusting the schema, since positional `param_bindings` and the document shape are both easy to get wrong. The stored header came back as a `varSpec` node keyed by the secret's ID, the signing binding resolved to the secret's name, every `dropped_content` was empty, and a second plan was empty.

**A name longer than 1024 bytes is now rejected at plan time**, where the API's cap used to surface part way through an apply. The limit is on bytes, and so is the validator, so the two agree on names using multi-byte characters.

---

> [!NOTE]
> **Low Risk**
> Changes are plan-time validation and documentation/examples only; secret create/update/rotate behavior is unchanged aside from earlier rejection of oversized names.
> 
> **Overview**
> **`incident_secret` names longer than 1024 bytes now fail at plan time** instead of during apply. The provider adds a byte-based `LengthAtMost` validator and documents the limit on `name`, including multi-byte character edge cases covered by new tests.
> 
> **Registry docs now show how to consume a secret in a `webhook.send` workflow step.** A second example (`resource-webhook-step.tf`) demonstrates both integration paths: embedding the secret ID in a header via `incident_rich_text` with `plain_single_line_with_secrets` (runtime substitution), and passing the secret ID directly as the step signing secret. A custom `templates/resources/secret.md.tmpl` splits “manage a secret” vs “use in webhook” like other multi-example resources; CHANGELOG reflects both changes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f91f568038cdfac11a4e034577619dc4e6f04c5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


benfdking (approved):
:chefskiss:

PR #572, opened 2026-09-09
Support moving state when a resource is renamed

benfdking:
**1 of 6 in the v7 stack.** Opened as a draft: the release has two open questions on it, listed at the bottom.

v7 promotes the beta resources to the names they will keep, and a rename is a change of resource type as far as Terraform is concerned. Terraform will not move state across types on its own — it cannot know whether the two schemas are compatible, and guessing loses somebody's state — so it asks the provider through the `MoveResourceState` RPC, and errors out if the target resource has not implemented it.

This adds the mover every one of those renames wants. The schema either side of a rename is the same one, so the move is a copy: the target ends up holding exactly what the source held, and a plan run straight afterwards reports no changes. That turns each rename into a `moved` block in a pull request:

```hcl
moved {
  from = incident_schedule_beta.primary
  to   = incident_schedule.primary
}
```

rather than a `terraform state mv` per resource, which is the difference between a migration somebody can review and one they run by hand against production state.

A mover that does not recognise its source returns no state and no diagnostics, which the framework reads as "skipped": it tries the next mover, and if none take the move it reports an error naming both types. That is a better error than any we could write, so recognising the source is all this has to do. It is deliberately strict about the schema version — writing state we do not understand into the target is worse than refusing, because a practitioner can undo a refusal.

Nothing consumes this yet; the promotions that do are the rest of the stack. The table in the test lists all five renames so the mover is known to accept each name before the resource claiming it lands.

## Open questions for the release

1. **Do the old resources really go, or become `_legacy` for one major?** Reusing the plain names means anyone who upgrades before migrating has state under `incident_schedule` that v7 cannot read, refresh or forget — `terraform state rm` plus a fresh import is the only way out. The guide in PR 2 has a recovery section, but a `_legacy` name for one major would mean nobody hits it.
2. **`incident_schedule_replica` needs a layer ID the new rotation resource cannot give it.** Detail in the schedules PR.

### DIFF
--- a/internal/provider/state_mover.go
+++ b/internal/provider/state_mover.go
@@ -1,0 +1,91 @@
+package provider
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+)
+
+// renameStateMover moves state from a resource that has been renamed, where the
+// schema either side of the rename is the same one.
+//
+// Terraform asks the provider to move state whenever a `moved` block changes a
+// resource's type, because it cannot know whether the two schemas are
+// compatible - and getting that wrong loses somebody's state. A rename is the
+// case where they are identical, so the move is a copy: whatever the source
+// resource held is what the target should hold, and a `terraform plan` straight
+// afterwards reports no changes.
+//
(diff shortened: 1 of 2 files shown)


benfdking:
Superseded: the state movers landed on master in #580, and #583 promoted the beta resources on top of them. Closing so the migration guide in #573 can sit directly on master.

PR #590, opened 2026-09-14
Accept when_alert_joins_group on alert routes that do not group

chill:
Teams are getting their own grouping preference, which groups a team's alerts on every alert route whether or not the route groups itself (incident-io/core#73485). A route with `grouping_config.default.enabled = false` therefore still needs to say how to escalate when an alert joins a group, and the v3 API now accepts and returns `escalation_config.when_alert_joins_group` for every route.

The provider tied that block to grouping in two places. The validator rejected a plan that set it with grouping off, and the plan modifier planned it back to null on the assumption the API would return null. Both go: the block is accepted whether or not the route groups, and is planned like any other v3 computed attribute, null in v2 mode, prior state when known, otherwise left for the API to compute.

This wants a release before `feature-alert-grouping-for-teams` reaches every organisation. 6,072 routes across 1,797 organisations have grouping off with a stored escalation mode, and any of them under Terraform would show a diff on every plan once their organisation is flagged on.

---

> [!NOTE]
> **Medium Risk**
> Changes alert-route escalation planning and refresh for many existing routes; wrong logic could cause noisy plans or drift in `when_alert_joins_group`, but scope is limited to that attribute and validation is relaxed rather than tightened.
> 
> **Overview**
> **Allows `escalation_config.when_alert_joins_group` on v3 `incident_alert_route` resources even when `grouping_config.default.enabled` is false**, matching team-level grouping that can group alerts on any route.
> 
> The provider no longer rejects that block at plan time when route grouping is off, and the plan modifier stops forcing it to null based on `grouping_enabled`. Planning now treats it like other v3 computed fields: null in v2, carry prior state when known, and leave unknown when creating, entering v3, or turning route grouping on. On read, when the API omits the mode (common before org-wide team grouping), **`FromAPIV3WithPlan` keeps the planned or last-known value** instead of clearing state and causing perpetual diffs.
> 
> Docs and CHANGELOG describe the behavior; unit tests cover the updated plan logic and API read-back.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 10dbe34aa78cf2622547038fae39534a3259a04e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/provider/alert_route_plan_modifiers.go
```diff
@@ -117,7 +101,7 @@ func autoRelateGroupedAlertsAction(configNull, planV3, planKnown, stateNull bool
 type whenAlertJoinsGroupPlanModifier struct{}
 
 func (whenAlertJoinsGroupPlanModifier) Description(context.Context) string {
-	return "Plans when_alert_joins_group as null when it does not apply (grouping_config unset, or grouping disabled), otherwise uses prior state when known."
+	return "Plans when_alert_joins_group as null when grouping_config is unset, otherwise uses prior state when known."
```
sbmooc: I don't think that this description is true? grouping_config can be unset, but we still maintain escalation config because of team preferences

sbmooc (approved):

PR #605, opened 2026-09-21
v6: fix plan failing on an alert route with incident creation disabled

### LINKED ISSUE #597: Plan fails with "Value Conversion Error" on alert route with incident creation disabled since v6.12.0
Provider version used: `v6.12.0`, `v6.13.0` (also reproduced by reading the code of `v7.0.0`). Works with `v6.11.0`.

I tried to update the provider from `v6.3.0`. Since `v6.12.0` the `tf plan` fails during refresh for one of our alert routes with the following message:

```
Error: Value Conversion Error

  with incident_alert_route.this,
  on alert-route.tf line 12, in resource "incident_alert_route" "this":
  12: resource "incident_alert_route" "this" {

An unexpected error was encountered while verifying an attribute value
matched its expected type to prevent unexpected behavior or panics. This is
always an error in the provider. Please report the following to the
provider developer:

Expected framework type from provider logic:
types.ObjectType["literal":jsontypes.NormalizedJSONOrStringType,
"reference":basetypes.StringType] / underlying type:
tftypes.Object["literal":tftypes.String, "reference":tftypes.String]
Received framework type from provider logic: types.ObjectType[] /
underlying type: tftypes.Object[]
Path: incident_template.name.value
```

The route is a page-only route with `incident_config.enabled = false`. Because incident creation is disabled, the API does not store the template and returns `name` with `autogenerated = false` and no binding. A template with a bound value fails the `apply` with `Provider produced inconsistent result after apply` (the API reads back empty), so the route ended up with the minimal template below. That shape applied fine on `v6.11.0` and planned with no changes, but cannot be read anymore since `v6.12.0`.

I bisected with the same configuration and state: `v6.11.0` plans clean, `v6.12.0` and `v6.13.0` fail. Other alert routes that still hold a bound name value are not affected.

```hcl
resource "incident_alert_route" "this" {
  name       = "Escalate to DEV"
  enabled    = true
  is_private = false

  alert_sources = [{
    alert_source_id  = "<ID_1>"
    condition_groups = []
  }]

  condition_groups = [{
    conditions = [
      {
        operation = "contains"
        subject   = "alert.title"
        param_bindings = [{
          value = { literal = "landing/" }
        }]
      },
    ]
  }]

  expressions    = []
  channel_config = []

  escalation_config = {
    auto_cancel_escalations = true
    escalation_targets = [{
      escalation_paths = {
        array_value = [{ literal = "<ID_2>" }]
      }
    }]
  }

  incident_config = {
    auto_decline_enabled    = false
    enabled                 = false
    condition_groups        = []
    defer_time_seconds      = 0
    grouping_keys           = []
    grouping_window_seconds = 0
  }

  incident_template = {
    custom_fields = []
    name = {
      autogenerated = false
    }
    summary = {
      autogenerated = false
    }
  }
}
```

As far as I can tell this comes from #552. In `alert_route_v2_model.go`, when the API returns `incident_template.name` without a binding and the prior state has a `name` block, the read path passes a zero-value `IncidentEngineParamBinding{}` on. The `objectOrNull` helper in `engine.go` should turn the untyped zero-value object into a typed null, but its nil check never fires because `ObjectValue.AttributeTypes` returns a freshly allocated map, which is empty but never nil. The v3 model passes `NullParamBinding()` at the same spot, which is probably why routes with `grouping_config` are not affected.

As a workaround we stay on `v6.11.0` for now.

benfdking:
Fixes #597. Targets `v6`, not `master`: the reporter is on the v6 line, and a fix on `master` only reaches them if they take v7, which is a breaking upgrade needing configuration rewrites. This gives them a v6.14.0. The same fix for v7 follows in its own PR.

A page-only alert route — `incident_config.enabled = false` with a minimal
`incident_template.name = { autogenerated = false }` — has failed every `terraform plan`
on refresh since **v6.12.0**:

```
Expected framework type: types.ObjectType["literal":jsontypes.NormalizedJSONOrStringType,
                                          "reference":basetypes.StringType]
Received framework type: types.ObjectType[]
Path: incident_template.name.value
```

The reporter's diagnosis was right in both halves, and this fixes both.

### The API returns no binding, so the read invents one with no type

Incident creation is off, so the API stores no template and answers with the name's
`autogenerated` flag and no binding at all. `FromAPIV2WithPlan` had nothing to build the
binding from and passed a bare `IncidentEngineParamBinding{}`, whose `Value` is a
`types.Object` carrying no attribute types. The framework checks that against the schema on
its way into state and refuses it — hence the error, at exactly that path.

The v3 model (`alert_route_v3_model.go`) and the incident template model
(`incident_template_v1_model.go`) already pass `NullParamBinding()` here, which types every
field. That is why routes with `grouping_config`, read through the v3 model, were never
affected. The v2 read now does the same.

### `objectOrNull` could never have caught it

```go
if obj.AttributeTypes(context.Background()) == nil {   // never true
```

`ObjectValue.AttributeTypes` copies into a freshly allocated map before returning it, so the
zero value's nil map arrives **empty rather than nil** and the `types.ObjectNull` branch was
unreachable. Its sibling `listOrNull` works only because `ListValue.ElementType` returns the
field itself — which is why `array_value` was fine and `value` was not. It now tests the
length, closing the same hole at the other call sites, where a zero-value binding reaching
`ToObject` would `panic` out of `types.ObjectValueMust` rather than produce a diagnostic.

### Regression

`objectOrNull` arrived in #552 (`b976a69`), first released in v6.12.0, when
`AlertRouteAutoGeneratedParamBinding.Value` moved from a `*IncidentEngineParamBindingValue`
(nil pointer → plain null, which is why v6.11.0 was clean) to a `types.Object`. The
reporter's bisect lands on exactly that release.

## Why it applies to v6 unchanged

The buggy code is identical at `v6.13.0`: the same three `IncidentEngineParamBinding{}` call
sites in `FromAPIV2WithPlan`, the same `objectOrNull`, and `incident_alert_route` already
accepts both the v2 and v3 schemas, so the v3 test has somewhere to go too. The only
adaptations are the changelog entry, which sits in the `v6` branch's `Unreleased` section
above `v6.13.0` alongside the #610 backport, and the module path in the unit tests, which is
`/v6`.

## Testing

Two unit tests, both red before the fix and green after, on this tree.
`TestAlertRouteV2TypesAnUnboundIncidentTemplate` reproduces the report through
`FromAPIV2WithPlan` and fails with the issue's exact mismatch:

```
value: expected types.ObjectType["literal":jsontypes.NormalizedJSONOrStringType, "reference":basetypes.StringType], got types.ObjectType[]
```

They assert on the binding's *type*, not its nullness — both spellings are null, and only
the typed one can be written to state, so nullness alone would pass against the bug.

```
go build ./... && go vet ./internal/...     ok
go test ./internal/provider/models/         ok
golangci-lint run ./...                     0 issues
```

The full CI run on this PR is green, including all three acceptance legs. (An earlier note
here called `TestAlertSourceAttributeBetaModifyPlanAcceptsValidBinding` a pre-existing
failure: that was my sandbox refusing `httptest` a port, not the test. It passes.)

Acceptance coverage: `TestAccIncidentAlertRouteResourceIncidentCreationDisabled` applies the
reporter's shape and asserts the refresh plans empty, and
`TestAccIncidentAlertRouteV3ResourceAutoGeneratedTemplate` covers the v3 route to the same
unbound binding. The v2 test has no import step: `ImportState` probes the v3 API first, so a
migrated organisation imports any route into the v3 schema and `ImportStateVerify` can never
match a v2 configuration.

---

> [!NOTE]
> **Low Risk**
> Targeted provider mapping fix for read/state conversion on alert routes; no auth or API behavior changes.
> 
> **Overview**
> Fixes a **v6.12.0 regression** where `terraform plan` failed on refresh for v2 `incident_alert_route` resources with **`incident_config.enabled = false`** and an **`incident_template.name`** (or summary) block with **no binding**—a common “page-only” route shape.
> 
> When the API returns no binding, **`FromAPIV2WithPlan`** now merges plan state using **`NullParamBinding()`** instead of an empty **`IncidentEngineParamBinding{}`**, so **`incident_template.name.value`** is a **typed null** object the framework accepts (matching v3 and other call sites). **`objectOrNull`** is corrected to treat a zero **`types.Object`** as untyped by checking **empty attribute types**, not a nil map, so the same class of conversion error is avoided elsewhere.
> 
> Adds **unit** coverage for unbound template round-trips and **`objectOrNull`**, plus **acceptance** tests for the v2 disabled-incident route and v3 autogenerated template refresh/import.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 170abb270ebb9aecc9fbaf8060cab59427cfa319. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


benfdking:
Retargeted from `master` to `v6` and force-pushed: the same three commits, rebased from `135f9c9` onto `v6`. The reporter on #597 is on the v6 line, so landing this on `master` only reached them through v7, a breaking upgrade. The fix is identical — the changelog entry moves into the v6 `Unreleased` section and the unit test's module path is `/v6`. The v7 side follows in its own PR.

Issue #595, opened 2026-09-16
owning_team_ids = [] causes "Provider produced inconsistent result after apply" on incident_alert_route create


yipclara:
### Provider version
`incident-io/incident ~> 6.0` (Terraform `>= 1.5.0`)

### Description

Creating an `incident_alert_route` with `owning_team_ids = []` (empty list) fails on apply with an inconsistent-result error. Every other alert route in our config has a non-empty `owning_team_ids` and applies fine — this only reproduces when the field is an empty list.

### Error

```
Error: Provider produced inconsistent result after apply

When applying changes to incident_alert_route.this, provider
"provider[\"registry.terraform.io/incident-io/incident\"]" produced an
unexpected new value: .owning_team_ids: was cty.SetValEmpty(cty.String),
but now null.

This is a bug in the provider, which should be reported in the provider's
own issue tracker.
```

### Minimal repro

```hcl
resource "incident_alert_route" "no_org_group_expiring_secrets" {
  name       = "No Org Group - Expiring Secrets"
  enabled    = true
  is_private = false

  owning_team_ids = []

  condition_groups = []
  expressions      = []

  alert_sources = [
    {
      alert_source_id  = "01M26FKC2KH26N82MQ1X3J0TQA"
      condition_groups = []
    },
  ]

  escalation_config = {
    auto_cancel_escalations = true
    escalation_targets      = []
  }

  incident_config = {
    enabled          = false
    condition_groups = []
  }

  grouping_config = {
    default = {
      enabled = false
    }
  }

  message_config = {
    destinations = [
      {
        slack_targets = {
          channel_visibility = "public"
          binding = {
            array_value = [
              { literal = "01M2GS4Y7KTTWKEX2SPY7PMNM4" },
            ]
          }
        }
        condition_groups = []
      },
    ]
  }
}
```

### Expected behavior
Apply succeeds, and `owning_team_ids` reads back as an empty set (unowned route), matching the planned value.

### Actual behavior
The API appears to drop the field entirely when it's empty and returns `null` instead of an empty set, so Terraform sees a mismatch between planned (`cty.SetValEmpty`) and actual (`null`) and aborts with "inconsistent result after apply."

### Workaround
Assign at least one team ID to `owning_team_ids`, or add `lifecycle { ignore_changes = [owning_team_ids] }` to the resource.

benfdking:
I've fixed this for v7 and will do a v6 patch.

PR #610, opened 2026-09-21
v6: keep an empty owning_team_ids empty on alert routes

benfdking:
Backport of #609 to the v6 line, for #595. Targets the new `v6` branch, which I cut at the `v6.13.0` tag — the last v6 release — since every release so far has been a point on `master` and there was no v6 line to land this on.

The reporter on #595 is pinned to `incident-io/incident ~> 6.0`, so the fix on `master` only reaches them if they take v7, which is a breaking upgrade needing configuration rewrites. This gives them a v6.14.0 instead.

## Why it applies cleanly

The alert route code at v6.13.0 is structurally identical: same model files, same `FromAPIV2WithPlan` / `FromAPIV3WithPlan` signatures, and `incident_alert_route` already accepted both the v2 and v3 schemas — which is how the reporter hit this with `grouping_config` set on a v6 provider.

The three code commits cherry-picked without conflict. Two adaptations:

- the module path in the new unit test is `/v6`, not `/v7`;
- the changelog entry sits under a fresh `## Unreleased` above `## v6.13.0`, and carries only this fix — not the unrelated v7 entries.

## Verification

- `go vet ./internal/...`, `gofmt` and `golangci-lint run ./...` (`0 issues`) all clean on the v6 tree.
- The new unit tests pass with the fix and fail without it: reverting just the two mappers to their v6.13.0 state gives `v2: owning_team_ids is null, want an empty set`, same for v3.
- `TestRequireJSONBody` and `TestAlertSourceAttributeBetaModifyPlanAcceptsValidBinding` already fail at a clean v6.13.0 checkout — I verified in a worktree at the tag, so they're not from this change.

Acceptance tests run on this PR. Note the acceptance jobs share a `concurrency` group keyed only by CLI and version, so runs across #607, #609 and this PR queue behind one another and a newer run cancels an older pending one.

---

> [!NOTE]
> **Low Risk**
> Localized state-mapping change for one optional attribute on alert routes, with thorough tests and no API or schema changes.
> 
> **Overview**
> Fixes **`incident_alert_route`** apply failures when **`owning_team_ids = []`**: the API omits that field for unowned routes, and the provider was writing **null** into state, which Terraform rejects as *Provider produced inconsistent result after apply* (#595).
> 
> Read mapping now goes through shared **`OwningTeamIDsToState`**, which uses the plan (or prior state on refresh) to tell an **explicit empty set** from an **attribute never set**—empty stays `[]` in state, omitted stays absent; imports with no plan still get null. **`FromAPIV2WithPlan`** and **`FromAPIV3WithPlan`** both use it.
> 
> Adds unit and acceptance coverage for create/refresh/update/omit on both schemas, documents the fix under **Unreleased** (backport of #609), and loosens an escalation-path acceptance regex so it matches either “empty array” or “empty slice” in the API error text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d655d60fceeb1bb53f2dad94e2a699b3be41a47. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


eespinosa (approved):

benfdking:
Added 7d655d6, which is unrelated to the backport itself — flagging it so it isn't mistaken for part of the fix.

The acceptance legs went red on `TestAccIncidentEscalationPathTeamIDs`, not on anything here: both backported tests passed on all three legs (`--- PASS: TestAccIncidentAlertRouteResourceEmptyOwningTeamIDs`, `--- PASS: TestAccIncidentAlertRouteV3ResourceEmptyOwningTeamIDs`).

That test pinned its `ExpectError` to `must set an empty slice or a list of Team`, and the API now answers `You must provide an empty array or a list of Team IDs` — a real 422 that just didn't match the pattern. The string is the API's, so the commit matches either spelling instead of today's text.

It only rots here because the v6-schema `incident_escalation_path` this covers was removed in v7 (#583), so the assertion doesn't exist on `master` and nothing has run it since `v6.13.0` was tagged. Worth knowing for the v6 line generally: 6 of its 42 `ExpectError` assertions match text that comes from the API rather than the provider, so they can drift the same way.

PR #616, opened 2026-09-22
Add the incident_pay_config resource and data source

benfdking:
Stacked on #615, which brings in the pay configs endpoints.

A pay config sets what someone is paid for being on call: a base rate, plus weekly rules and one-off rules that override it at particular times. This adds:

- The `incident_pay_config` resource, which manages a config with its rules as one thing. `weekly_rules` is ordered, since the first rule that covers a shift prices it; `one_off_rules` take precedence and may not overlap.
- The `incident_pay_config` data source, which looks a config up by `id` or `name`.

### How updates work

The API splits a config into three parts with their own endpoints: the config's attributes (whose update leaves rules alone), weekly rules, and one-off rules. Update reconciles each part and sends only what changed. That matters beyond economy: editing a config that a published report priced against needs the `schedule_pay_configs.update_published` scope, and adding a holiday to it should not.

Rules are matched to the API **by position**. Weekly rules are evaluated in order and the API cannot reorder them, so rewriting positions in place, appending, and trimming from the end are the only moves that keep the configured order. They are also what the plan promises once each position's ID is carried over from state. One-off rules use the same scheme, with removals before updates and additions, so swapping one holiday for an overlapping one never holds both at once (the API rejects overlaps on every write).

### Two framework workarounds

- `UseStateForUnknown` on a nested computed attribute copies the *null* it is handed for a list element beyond the end of state, so a newly appended rule's ID was planned as null and the apply failed the consistency check. `useStateForUnknownIfSet` (in `plan_modifiers.go`) leaves such a value unknown.
- `timetypes.RFC3339` compares timestamps by formatted string, offset included, so a window written as `+01:00` and read back in UTC was a perpetual diff. `timestamptypes.Instant` compares the moment instead. It is only used for the one-off rule window; the read-only timestamps keep `timetypes.RFC3339`.

### Testing

- `TestIncidentPayConfigResourceLifecycle` drives the resource through Terraform against an in-memory fake of the API (no credentials needed) and asserts the exact writes each step makes: attributes-only changes hit only the config endpoint, a rule change hits only that rule, removals land before additions, and each position keeps its ID.
- `TestAccIncidentPayConfigResource` runs the same steps against a real account. **I could not run it**: there is no API key in this environment and no local core. The fake mirrors core's `MutatePayConfig`/`rules.go` behaviour (append order, overlap check, UTC timestamps), but a run against a real account before merging would be worth it. The key needs the `pay_configs_editor` role.
- Plan-time validation and model conversions have their own unit tests.
- `golangci-lint` did not run locally (the installed binary was built with an older Go than the repo targets), so CI is the first lint pass.

The pre-existing `TestPolicyBlocksCoverEveryPolicyType` failure is from #615 and is unrelated.

---

> [!NOTE]
> **Medium Risk**
> New pay-configuration surface with position-based rule reconciliation; incorrect updates could change on-call compensation, though scope is additive and heavily tested.
> 
> **Overview**
> Adds **`incident_pay_config`** and **`incident_pay_config`** (data source) so Terraform can manage on-call pay rates: base rate, ordered **`weekly_rules`**, and non-overlapping **`one_off_rules`**, with import and registry docs/examples.
> 
> The resource **updates in three API slices**—config fields only when they change, then weekly rules, then one-off rules—reconciling rule lists **by list position** (update in place, append, trim) so rule-only changes avoid the **`schedule_pay_configs.update_published`** scope and one-off swaps delete before update to satisfy overlap checks. Plan-time validation mirrors API rules (IANA timezone, ISO currency, clock times, one-off overlap).
> 
> Supporting changes: **`timestamptypes.Instant`** for one-off windows (same instant across UTC vs offset), **`useStateForUnknownIfSet`** for nested rule IDs on append, shared **`PayConfigModel`**, and lifecycle tests against an in-memory fake API plus plan-only validation tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d7174e212322e844f23bca4e3e23806cd4e7601. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


internal/provider/incident_pay_config_resource.go
```diff
@@ -0,0 +1,626 @@
...
+rule that should win first. `+"`one_off_rules`"+` take precedence over every weekly rule and may
+not overlap each other, so their order has no effect on pricing. Any time no rule covers is paid
+at `+"`base_rate_cents`"+`.
+
+Terraform creates the config as a draft, which becomes visible to everyone in the organisation
+once a report that prices against it is published. Editing a config after that changes the
+explanation of pay someone has already been sent, so it additionally needs the
+`+"`schedule_pay_configs.update_published`"+` scope on the API key, which the `+"`pay_configs_editor`"+`
+role does not carry. The provider only sends what changed, so a plan that touches nothing but
+the rules never asks to update the config itself.`),
```
joladev: I think the permission is required for every change, not just rules

joladev: and changing rules requires update_published

joladev: pay_configs_editor has the permission

examples/resources/incident_pay_config/resource.tf
```diff
@@ -0,0 +1,53 @@
...
+    {
+      weekdays   = ["saturday", "sunday"]
+      start_time = "00:00"
+      end_time   = "00:00" # equal to start_time means the whole day
+      rate_cents = 1500
+    },
+    {
+      weekdays   = ["monday", "tuesday", "wednesday", "thursday", "friday"]
+      start_time = "18:00"
+      end_time   = "09:00" # runs past midnight into the next morning
```
joladev: a rule start_time->end_time has to be a single calendar day, this would just not match. to do over midnight you need two rules

joladev (approved):
minor stuff

